PR #147, opened 2024-09-26
Better install experience

mrbojangles3:


pkg/fab/flatcar/iso/build.go
```diff
@@ -12,33 +12,216 @@
...
 
-// Build builds the Control Node ISO only, based on the pre-built control-instal bundle, not generic
+// Copies a file from the local directory to the newly created filesystem, does not rename files.
+func copyFile(dstPath string, srcPath string, destination filesystem.FileSystem) error {
+	//slog.Debug("CopyFile Entry", "DstPath", dstPath, "SrcPath", srcPath, "Destination Filesystem", destination.Label())
+	//src, err := os.Open(srcPath)
+	// here goes all the memory...
+	src, err := os.ReadFile(srcPath)
+	if err != nil {
+		slog.Error("CopyFile Error Opening Source Path", "SourcePath", srcPath, "Error:", err.Error())
```
Frostman, line 40: We shouldn't ignore such errors, plus let's use fmt.Errorf to wrap errors, please. I wrote you in slack to have chat about it.

pkg/fab/flatcar/iso/build.go
```diff
@@ -12,33 +12,216 @@
...
+		espPartitionStart   int64 = 2048
+		espPartitionSectors int64 = espSize / blkSize                             // 1024000 sectors
+		espPartitionEnd     int64 = espPartitionSectors + (espPartitionStart - 1) // 1026047
+		oemPartitionStart   int64 = espPartitionEnd + 1                           // 1026048
+		oemPartitionSectors int64 = oemSize / blkSize                             // 2097152 sectors
+		oemPartitionEnd     int64 = oemPartitionSectors + (oemPartitionStart - 1) // 3123199
+	)
+
+	// create a disk image
+	diskLogan, err := diskfs.Create(diskImg, diskSize, diskfs.Raw, diskfs.SectorSizeDefault)
```
Frostman, line 135: "diskLogan"

pkg/fab/flatcar/iso/build.go
```diff
@@ -12,33 +12,216 @@
...
+		slog.Error("Error copying tree", "Error", err.Error())
+		return err
+	}
+	err = copyTree(workdir, "/boot", espFs)
+	if err != nil {
+		slog.Error("Error copying tree", "Error", err.Error())
+		return err
+	}
+
+	err = copyFile("/", workdir+"/flatcar_production_pxe_image.cpio.gz", espFs)
```
Frostman, line 198: each error should be handled

Frostman (changes requested):
Added some initial comments

pkg/fab/ctrl_os_butane.tmpl.yaml
```diff
@@ -176,3 +176,21 @@ storage:
...
       group:
         name: core
+systemd:
+  units:
+    - name: fabric-install.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description="Firstboot installation program for HedgeHog Fabric"
+        ConditionFirstBoot=yes
```
Frostman, line 186: I'm wondering if installer failed on first attempt for some reason, do we want to re-run it automatically? Would first boot handle it?

pkg/fab/flatcar/iso/build.go
mrbojangles3: lol, ah yeah. Forgot to clean that up.

pkg/fab/ctrl_os_butane.tmpl.yaml
mrbojangles3: While it is technically possible to re-run if the installer fails for some reason it would be terrible UX. 
We can change this unit file to run conditionally on the existence of a file for example `/opt/hedgehog/install-done` and if the install fails for some reason we can instruct the user to delete that file then re-run the unit file with `systemctl start fabric-install.service` or some nicer shell script with that in it.

Frostman:
Installer ISO build is merged into hhfab-v2 branch as https://github.com/githedgehog/fabricator/pull/148/commits/65afe2c04fb683b0408ffd96c905c644adeba1ac

Frostman:
Merged as part of other PRs

PR #230, opened 2024-11-25
Minor improvs to hhnet script

edipascale:
* set down interface before enslaving it, to prevent errors if it is already up
* limit the attempts in get_ip to 30 to avoid stalling forever

### DIFF
--- a/pkg/hhfab/hhnet.sh
+++ b/pkg/hhfab/hhnet.sh
@@ -26,6 +26,8 @@ function setup_bond() {
     sudo ip l a "$bond_name" type bond miimon 100 mode 802.3ad
 
     for iface in "${@:2}"; do
+        # cannot enslave interface if it is up
+        sudo ip l s "$iface" down 2> /dev/null || true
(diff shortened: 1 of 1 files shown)


pkg/hhfab/hhnet.sh
```diff
@@ -44,13 +46,19 @@ function setup_vlan() {
 function get_ip() {
     local iface_name=$1
     local ip=""
+    local max_attempts=30
```
Frostman, line 49: let's set it to a minute just to be on a safe side, thank you

Frostman (changes requested):
thx, it totally makes sense to down interface just in case and not wait forever for IP from DHCP

pkg/hhfab/hhnet.sh
edipascale: done, thanks!

mrbojangles3:
After this passes CI, it looks good to me.

Frostman (approved):

PR #236, opened 2024-12-02
add force-reset script in vlab

edipascale:
useful for automated testing. also fix a typo in hhfab

### DIFF
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -647,7 +647,7 @@ func Run(ctx context.Context) error {
 								Value:   true,
 							},
 							&cli.BoolFlag{
-								Name:    "force-clenup",
+								Name:    "force-cleanup",
 								Aliases: []string{"f"},
 								Usage:   "start with removing all existing VPCs and VPCAttachments",
 							},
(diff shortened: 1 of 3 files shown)


cmd/hhfab/main.go
```diff
@@ -647,7 +647,7 @@ func Run(ctx context.Context) error {
 								Value:   true,
 							},
 							&cli.BoolFlag{
-								Name:    "force-clenup",
+								Name:    "force-cleanup",
```
Frostman, line 650: ouch, nice catch, could you please make a PR only with fixing the flag name? thx

pkg/hhfab/vlabrunner.go
```diff
@@ -640,6 +643,21 @@ func (c *Config) vmPostProcess(ctx context.Context, vlab *VLAB, d *artificer.Dow
 			return fmt.Errorf("trying toolbox: %w", err)
 		}
 	} else if vm.Type == VMTypeControl {
+		slog.Debug("Installing force_reset", "vm", vm.Name, "type", vm.Type)
```
Frostman, line 646: it shouldn't be a bash script, the fabric cli is a good place for it - https://github.com/githedgehog/fabric/tree/master/pkg/hhfctl

Frostman (changes requested):

pkg/hhfab/vlabrunner.go
edipascale: ack - closing this, will look into implementing it in fabric directly

PR #256, opened 2024-12-09
add hhfab switch reinstall and power subcommands

pau-hedgehog:
Adds `switch power|reinstall` suboptions to `vlab` in `hhfab`.

-  `vlab switch power` enables power management via external PDU servers. For example:

```
hhfab vlab switch --verbose power --yes --all ON
hhfab vlab switch --verbose power --yes --name leaf-01 CYCLE
```

A new annotation structure (`power:`) is supported:

```
core@control-1 ~ $ kubectl get switch leaf-01 -o=jsonpath='{.metadata.annotations}' | jq
{
  "power.hhfab.githedgehog.com/psu1": "http://192.168.101.10/outlet/1",
}
```

The pdu credentials are configured in a YAML file in the same folder as `fab.yaml` with the following structure:
```
$ cat .pdu.yaml 
pdus:
  192.168.101.10:
    user: user10
    password: password10
  192.168.101.11:
    user: user11
    password: password11
```

-  `vlab switch reinstall` enables automated NOS reinstall of switches. For example:

```
hhfab-power vlab switch -v reinstall --yes --all --mode hard-reset
hhfab-power vlab switch -v reinstall --yes --name leaf-01
```

The default mode (`reboot`) logs into the switch console, reboots and goes into ONIE NOS install mode to reboot once more and let the switch ready to be managed by the Control node.

The `hard-reset` mode relies on the power CYCLE operation to cut power briefly for the given switch(es).

Depends On https://github.com/githedgehog/fabric/pull/689


pkg/pdu-mgt/go.mod
```diff
@@ -0,0 +1,3 @@
+module go.githedgehog.com/fabricator/pkg/pdu-mgt
```
Frostman, line 1: There is no need for a separate module. It only adds some complications. Could you please move the code to somewhere under pkg/hhfab like pkg/hhfab/pdu

pkg/pdu-mgt/utils/utils.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return id
+}
+
+func GetPDUIPFromURL(url string) string {
+	parts := strings.Split(url, "/")
+	return parts[2]
+}
+
+func ConfirmActions(action string, outlets map[string]string) bool {
```
Frostman, line 67: There are a few places where we're using --yes flag in such cases in a CLI, I think it'll be better to keep that approach consistent across CLIs. Last example is https://github.com/githedgehog/fabric/pull/690

pkg/pdu-mgt/utils/utils.go
```diff
@@ -0,0 +1,81 @@
...
+	id, err := strconv.Atoi(outletID)
+	if err != nil {
+		log.Fatalf("Error extracting outlet ID: %v", err)
+	}
+	return id
+}
+
+func GetPDUIPFromURL(url string) string {
+	parts := strings.Split(url, "/")
+	return parts[2]
```
Frostman, line 64: it'll panic if len(parts) < 0, so it needs a check

pkg/pdu-mgt/utils/utils.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return outlets, uniqueIPs, nil
+}
+
+func ExtractOutletID(url string) int {
+	parts := strings.Split(url, "/")
+	outletID := parts[len(parts)-1]
+	id, err := strconv.Atoi(outletID)
+	if err != nil {
+		log.Fatalf("Error extracting outlet ID: %v", err)
```
Frostman, line 57: error should be returned or handled

cmd/hhfab/main.go
```diff
@@ -821,6 +821,73 @@ func Run(ctx context.Context) error {
...
+						Before: before(false),
+						Subcommands: []*cli.Command{
+							{
+								Name:  "reinstall",
+								Usage: "reinstall one or all switches",
+								Flags: []cli.Flag{
+									&cli.StringFlag{
+										Name:    "name",
+										Aliases: []string{"n"},
+										Usage:   "name of the switch to reinstall, or use '--all' for all switches",
```
Frostman, line 837: I think it'll be better to have a separate bool flag `--all` so it's an explicit flag

cmd/hhfab/main.go
```diff
@@ -821,6 +821,73 @@ func Run(ctx context.Context) error {
...
+									},
+								},
+								UsageText: "hhfab vlab switch power [--name <switchName>|--all] <action>",
+								Action: func(c *cli.Context) error {
+									switchName := c.String("name")
+									if switchName == "" {
+										return fmt.Errorf("missing required flag: --name/-n") //nolint:goerr113
+									}
+
+									if c.NArg() < 1 {
```
Frostman, line 872: I think stricter check (!=1) would allow to avoid unexpected behaviour

cmd/hhfab/main.go
```diff
@@ -821,6 +821,73 @@ func Run(ctx context.Context) error {
...
+										return fmt.Errorf("missing power action (ON, OFF, CYCLE)") //nolint:goerr113
+									}
+
+									powerAction := c.Args().First()
```
Frostman, line 876: maybe wrap with strings.ToUpper just in case

Frostman (changes requested):

pkg/pdu-mgt/go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/joho/godotenv v1.5.1 h1:7eLL/+HRGLY0ldzfGMeQkb7vMd0as4CfYvUVzLqw0N0=
```
Frostman, line 1: It doesn't look like it's used anywhere

pkg/pdu-mgt/go.mod
pau-hedgehog: Thanks for the hint. Yes gave me more than one headache.

Addressed

pkg/pdu-mgt/utils/utils.go
pau-hedgehog: Addressed taking the approach from fabric

pkg/pdu-mgt/utils/utils.go
pau-hedgehog: addressed

pkg/pdu-mgt/utils/utils.go
pau-hedgehog: addressed

pkg/pdu-mgt/go.sum
pau-hedgehog: removed as it's been integrated in hhfab

cmd/hhfab/main.go
pau-hedgehog: addressed

cmd/hhfab/main.go
pau-hedgehog: addressed

cmd/hhfab/main.go
pau-hedgehog: ...and fixed now for power operations

pau-hedgehog:
After discussing, the default behavior needs to change and reinstall must leave the switches in the ONIE install loop.

The initial behavior will be preserved with a flag `--wait-ready` that will be relayed to the GRUB script.

pau-hedgehog:
> After discussing, the default behavior needs to change and reinstall must leave the switches in the ONIE install loop.
> 
> This way one can trigger the reinstall and then do the `vlab up`. Once the control node is ready, the switches will be finally reinstalled.
> 
> The initial behavior will be preserved with a flag `--wait-ready` that will be relayed to the GRUB script.

I tried this but hhfab requires vlab up:
```
ubuntu@env-1:~/hhfab$ ./hhfab-pow-reinst vlab switch reinstall --all --yes 
07:37:04 INF Hedgehog Fabricator version=v0.32.1-21-gec342fd8-dirty-a72344
Enter username: admin
Enter password: 
07:37:10 INF Wiring hydrated successfully mode=if-not-present
07:37:10 ERR reinstall failed: preparing VLAB: VLAB directory does not exist: "/home/ubuntu/hhfab/vlab"
```

Shall we go with the original approach, @Frostman ?

Frostman (approved):

PR #259, opened 2024-12-09
Remove hard-coded values, except when there are no arguments supplied

mrbojangles3:


Frostman (changes requested):
It breaks the UX of minimal config to make it work, e.g., if you specify the number of leafs, the number of spines would still be 0 as well as no test servers. I think a sane number of test servers should be a default. Additionally, it would potentially break a CI as a default topo changes.

mrbojangles3:
Closing after chat with Sergei, will re-open a new PR based on that conversation.

PR #272, opened 2024-12-13
initial version of switch reinstall

pau-hedgehog:
Adds `reinstall` suboption to `vlab switch` in `hhfab` to enable the NOS reinstall of one or all switches. For example:

```
hhfab vlab switch reinstall --yes --all --mode soft-reset
hhfab vlab switch --verbose reinstall  --yes -n leaf-01
```

Three modes of operations are available:
- reload: current default. Connects to the switch via serial and reloads (requires username/password)
- soft reset: queries the power reset from the Agent API
- hard-reset: will be the default option
  - queries the PDU API and issues a power cycle (pending)
  - Depends On https://github.com/githedgehog/fabricator/pull/256

### DIFF
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -834,6 +834,39 @@ func Run(ctx context.Context) error {
 							return nil
 						},
 					},
+					{
+						Name:   "switch",
+						Usage:  "manage switch reinstall or power",
+						Flags:  append(defaultFlags, accessNameFlag),
+						Before: before(false),
+						Subcommands: []*cli.Command{
+							{
+								Name:  "reinstall",
+								Usage: "reinstall one or all switches",
+								Flags: []cli.Flag{
+									&cli.StringFlag{
+										Name:    "name",
+										Aliases: []string{"n"},
+										Usage:   "name of the switch to reinstall, or use '--all' for all switches",
+									},
+								},
+								Action: func(c *cli.Context) error {
+									switchName := c.String("name")
+									if switchName == "" {
+										return fmt.Errorf("missing switch name or --all") //nolint:goerr113
+									}
+
+									err := hhfab.DoSwitchReinstall(ctx, workDir, cacheDir, switchName)
+									if err != nil {
+										return fmt.Errorf("reinstall failed: %w", err)
+									}
+
+									return nil
+								},
+								HelpName: "hhfab vlab switch reinstall",
+							},
+						},
+					},
 				},
 			},
 			{
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -23,6 +24,7 @@ import (
 	"go.githedgehog.com/fabricator/pkg/fab/recipe"
 	"go.githedgehog.com/fabricator/pkg/hhfab"
 	"go.githedgehog.com/fabricator/pkg/version"
+	"golang.org/x/term"
 	"gopkg.in/natefinch/lumberjack.v2"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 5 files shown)


cmd/hhfab/main.go
```diff
@@ -821,6 +821,39 @@ func Run(ctx context.Context) error {
...
+										Usage:   "name of the switch to reinstall, or use '--all' for all switches",
+									},
+								},
+								Action: func(c *cli.Context) error {
+									switchName := c.String("name")
+									if switchName == "" {
+										return fmt.Errorf("missing switch name or --all") //nolint:goerr113
+									}
+
+									err := hhfab.DoSwitchReinstall(ctx, workDir, cacheDir, switchName, verbose)
```
edipascale, line 846: nit: `if err := ...; err != nil {...`

pkg/hhfab/vlabhelpers.go
```diff
@@ -236,3 +238,92 @@ type VLABAccessInfo struct {
...
+	var mu sync.Mutex
+
+	for _, sw := range targets {
+		wg.Add(1)
+		go func(sw wiringapi.Switch) {
+			defer wg.Done()
+
+			cmdName := VLABCmdGrubSelect
+			var cmd *exec.Cmd
+			cmd = &exec.Cmd{}
```
edipascale, line 279: nit: why not initialize the variable directly with `cmd := &exec.Cmd{}`?

pkg/hhfab/vlabhelpers.go
```diff
@@ -236,3 +238,92 @@ type VLABAccessInfo struct {
...
+			cmd.Dir = c.WorkDir
+			cmd.Stdin = os.Stdin
+			args := []string{sw.Name}
+			if verbose {
+				byobuCmd := fmt.Sprintf("byobu new-window -d -n %s '%s %s'", sw.Name, cmdName, strings.Join(args, " "))
```
edipascale, line 284: I don't think we can rely on the fact that byobu is installed in general, I'd just run the command

pkg/hhfab/vlabhelpers.go
```diff
@@ -236,3 +238,92 @@ type VLABAccessInfo struct {
...
+			if err := cmd.Run(); err != nil {
+				mu.Lock()
+				errs = append(errs, fmt.Errorf("\n%s failed for switch %s: %w", cmdName, sw.Name, err))
+				mu.Unlock()
+			}
+		}(sw)
+	}
+
+	for _, sw := range targets {
+		slog.Info("Executing soft-reset on", "switch", sw.Name+"...")
```
edipascale, line 304: should we not do the power cycle before starting the expect scripts? and what is the purpose of the sleep at the end of the loop?

pkg/hhfab/vlabhelpers.go
```diff
@@ -236,3 +238,92 @@ type VLABAccessInfo struct {
...
+				cmd.Stdout = os.Stdout
+				cmd.Stderr = nil
+			} else {
+				cmd = exec.CommandContext(ctx, cmdName, args...)
+				cmd.Stdout = nil
+				cmd.Stderr = os.Stderr // expect messages logged to stderr to show progress
+			}
+
+			slog.Info("Running", "cmd", strings.Join(append([]string{cmdName}, sw.Name+"..."), " "))
+			if err := cmd.Run(); err != nil {
```
edipascale, line 295: wonder whether we should add some sort of timeout to make sure this terminates, just as a safety mechanism

pau-hedgehog:
@Frostman, the `expect` requirement is breaking the `vlab` stages in this PR:

https://github.com/githedgehog/fabricator/actions/runs/12398209633/job/34610743571#step:7:81

How can we add it to the runner?

pkg/hhfab/vlabhelpers.go
pau-hedgehog: Yes, I don't want to add the dependency. But otherwise you would be getting the output of N commands in the same window, which is ugly. It was convenient to debug in the first iterations of the expect script integration, though

cmd/hhfab/main.go
pau-hedgehog: Addressed

pkg/hhfab/vlabhelpers.go
pau-hedgehog: Done

pkg/hhfab/vlabhelpers.go
pau-hedgehog: The order works as of now, we could change it. The sleep is to try to increase the odds of agent power reset working. We can remove if not needed for the PDU API

pkg/hhfab/vlabhelpers.go
pau-hedgehog: Timeout handling added

pau-hedgehog:
Closing in favor of https://github.com/githedgehog/fabricator/pull/256

PR #280, opened 2024-12-17
ISO build speed up and vlab recreate

Frostman:


pkg/fab/recipe/control_build_usb.go
```diff
@@ -300,7 +296,7 @@ func diskFSCopyFile(dstPath string, srcPath string, destination filesystem.FileS
 	if dstPath == "/" {
 		dstPath = filepath.Join("/", filepath.Base(srcPath))
 	}
-	dest, err := destination.OpenFile(dstPath, os.O_CREATE|os.O_RDWR|os.O_SYNC)
+	dest, err := destination.OpenFile(dstPath, os.O_CREATE|os.O_RDWR|os.O_TRUNC)
```
mrbojangles3, line 299: I can't remember if FAT filesystem, supports TRUNC via go-diskfs

Frostman: TRUNC is implicit behaviour is APPEND isn't specified and go-diskfs most probably only looking for RDWR and APPEND ignoring everything else

mrbojangles3: I agree on removing sync. It is probably a holdover from the debugging I was doing.

mrbojangles3 (approved):

PR #287, opened 2024-12-19
Update readme

mrbojangles3:


README.md
```diff
@@ -1,98 +1,178 @@
...
+* `hhfab init --dev --registry-repo 127.0.0.1`
+* `hhfab vlab gen`
+* `hhfab vlab up --mode iso`
 
-## Contributing
-// TODO(user): Add detailed information on how you would like others to contribute to this project
+### updating pods
 
-**NOTE:** Run `make help` for more information on all potential `make` targets
+* (TODO)
```
Frostman, line 176: KUBECONFIG env var set to target cluster that have access to the registry you push to (or you can use registry on control node in the upstream mode and push directly to it, in case of VLAB the registry port on control node is forwarded from the host as well) and run oras oci_repo="..." push patch

mrbojangles3: making a note from our conversation - pushing to the registry *on* the control node has limited utility. It is better in most cases to use upstream mode.

README.md
```diff
@@ -1,98 +1,178 @@
...
+
+### hhfab
+
+If the code you are changing deals with setting up or managing flatcar, hhfab
+will need to be instructed to pull packages from the local zot registry and not
+the ghcr. To do this, specify the repo pass the `--registry-repo
+127.0.0.1:30000` flag and argument to  `hhfab init` along with other flags.
+From there continue on with the `hhfab` commands. To get vlab running with
+local changes:
+* `hhfab init --dev --registry-repo 127.0.0.1`
```
pau-hedgehog, line 170: Won't defining localhost address end up in error when creating a control node VM in your local setup?

README.md
```diff
@@ -1,98 +1,178 @@
...
-kubectl apply -k config/samples/
 ```
+{
+  "log": {
+    "level": "debug"
+  },
+  "storage": {
+    "rootDirectory": "/tmp/zot"
+  },
+  "http": {
```
pau-hedgehog, line 70: Is plain HTTP valid for push/pull forl all of the operations in the justfile or other places where the registry is required?

README.md
```diff
@@ -1,98 +1,178 @@
...
-## Getting Started
+## Description 
+
+The components of this repository are distributed and versioned independently
+as OCI compliant artifacts.
+
+
+## Local Build Instructions
 
 ### Prerequisites
```
pau-hedgehog, line 15: missing just (requires 1.38.0?)

README.md
```diff
@@ -1,98 +1,178 @@
...
 
-```sh
-make install
-```
+Zot is an OCI package registry. Zot is used on your local system as a
+pull-through cache for all artifacts that are not being changed locally as part
+of development process.
 
-**Deploy the Manager to the cluster with the image specified by `IMG`:**
+[Zot installation
```
pau-hedgehog, line 47: Is there a reason why the _localreg just target cannot be used here?

README.md
pau-hedgehog, line 170: missing `:30000`

mrbojangles3: thank you for spotting this. adding it in my next commit

README.md
mrbojangles3: This is a good question, the steps in the justfile are aimed at getting zot setup in CI. These steps are very similar but since (at least IMO) a dev would have a persistent zot registry the instructions should go far enough to have persistence and a systemd unit file.

README.md
mrbojangles3: We are using plain HTTP in locally to avoid the self signed certificate hurdle. The plain http option is used in hhfab for pull, and just file for pushing. I am not aware of other uses of zot, am I missing one?

pau-hedgehog: My use case is maybe out of scope here. I was developing locally and pushing to env-1 to test different stuff. So I needed the artifacts required for hhfab and also the ones required during the control node installation available

mrbojangles3: In conversation with Sergei, He is planning on setting up zot inside the labs so the lab would have its own "local" registry.

pau-hedgehog:
@mrbojangles3 , mind the new DCO checks:

> To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

mrbojangles3:
Closing, deleting, and re-opening. I lost the ability to amend the signoffs.

PR #289, opened 2024-12-27
Enable LLDP on Server VMs

pau-hedgehog:
Enables LLDP on server VM interfaces to cross-check topology:

```
core@server-1 ~ $ networkctl -l lldp 
LINK   CHASSIS-ID        SYSTEM-NAME CAPS        PORT-ID   PORT-DESCRIPTION                                      
enp2s1 54:bf:64:ba:3d:c0 s5248-01    ....r...... Ethernet1 PC2 MCLAG server-1 server-1--mclag--s5248-01--s5248-02
enp2s2 3c:2c:30:66:f0:00 s5248-02    ....r...... Ethernet1 PC2 MCLAG server-1 server-1--mclag--s5248-01--s5248-02

Capability Flags:
r - Router

2 neighbors listed.
```

Also enables DHCP on the same interfaces

### DIFF
--- a/pkg/hhfab/vlab_butane.tmpl.yaml
+++ b/pkg/hhfab/vlab_butane.tmpl.yaml
@@ -56,12 +56,14 @@ storage:
           TOOLBOX_DOCKER_TAG=latest
           TOOLBOX_USER=root
 
-    - path: /etc/systemd/network/99-wildcard.network
+    - path: /etc/systemd/network/99-lldp.network
       mode: 0644
       contents:
         inline: |
           [Match]
-          Name=en*
+          Name=*
+          Type=ether
+          Kind=!vlan
 
           [Network]
           DHCP=yes
--- a/pkg/hhfab/vlab_butane.tmpl.yaml
+++ b/pkg/hhfab/vlab_butane.tmpl.yaml
@@ -55,3 +55,15 @@ storage:
           TOOLBOX_DOCKER_IMAGE=ghcr.io/githedgehog/toolbox
           TOOLBOX_DOCKER_TAG=latest
           TOOLBOX_USER=root
+
+    - path: /etc/systemd/network/99-wildcard.network
+      mode: 0644
+      contents:
+        inline: |
+          [Match]
+          Name=en*
+
+          [Network]
+          DHCP=yes
+          LLDP=yes
+          EmitLLDP=yes


hack/tools.just
```diff
@@ -136,8 +136,8 @@ test path="./...": gen _envtest _gcov2lcov
 #
 
 oci := ""
-skopeo_dest_insecure := if oci == "http" { "--dest-tls-verify=false" } else { "" }
-oras_insecure := if oci == "http" { "--plain-http" } else { "" }
+skopeo_dest_insecure := if oci == "http" { "" } else { "--dest-tls-verify=false" }
```
Frostman, line 139: the `else` branch is supposed to be "secure", what are you trying to achieve?

hack/tools.just
```diff
@@ -153,7 +153,7 @@ _helm-build name: _helm
   {{helm}} lint config/helm/{{name}}-{{version}}.tgz
 
 _helm-push name: _helm
-  {{helm}} push config/helm/{{name}}-{{version}}.tgz oci://{{oci_repo}}/{{oci_prefix}}/charts
+  {{helm}} push --insecure-skip-tls-verify config/helm/{{name}}-{{version}}.tgz oci://{{oci_repo}}/{{oci_prefix}}/charts
```
Frostman, line 156: same - default behaviour should be secure

pkg/hhfab/vlab_butane.tmpl.yaml
```diff
@@ -20,6 +20,28 @@ systemd:
     - name: locksmithd.service
       mask: true
 
+    - name: systemd-resolved.service
+      enabled: true
+      dropins:
+        - name: restart.conf
+          contents: |
+            [Service]
+            ExecStartPost=/bin/systemctl restart systemd-resolved
```
Frostman, line 29: Why? It looks kinda strange

pkg/hhfab/vlab_butane.tmpl.yaml
```diff
@@ -20,6 +20,28 @@ systemd:
...
+
+    - name: lldpd.service
```
Frostman, line 31: It would be better to enable LLDP through the systemd-networkd which is used for the interfaces configurations, we already have it for the mgmt interface - https://github.com/githedgehog/fabricator/blob/135b17c8fd155df192caad131e657f5ede98035e/pkg/fab/recipe/control_butane.tmpl.yaml#L106-L107 

I think something like but applied to all interfaces would be a good approach (together with DHCP, check the 99-default file (or similar name)

pkg/hhfab/vlab_butane.tmpl.yaml
```diff
@@ -55,3 +77,18 @@ storage:
           TOOLBOX_DOCKER_IMAGE=ghcr.io/githedgehog/toolbox
           TOOLBOX_DOCKER_TAG=latest
           TOOLBOX_USER=root
+
+    - path: /etc/systemd/resolved.conf.d/custom-dns.conf
```
Frostman, line 81: We shouldn't configure DNS by default as we're testing a DHCP server feature to propagate specified DNS servers. Check `--dns` flag for `hhfab vlab setup-vpcs`

pkg/hhfab/vlab_butane.tmpl.yaml
```diff
@@ -55,3 +77,18 @@ storage:
...
+      mode: 0644
+      contents:
+        inline: |
+          [Resolve]
+          DNS=8.8.8.8 8.8.4.4
+          FallbackDNS=1.1.1.1
+
+    - path: /etc/lldpd.conf
```
Frostman, line 89: Woudn't be required if using systemd-networkd LLDP

pkg/hhfab/vlab_butane.tmpl.yaml
```diff
@@ -20,6 +20,28 @@ systemd:
...
+
+    - name: lldpd.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=LLDP Daemon
+        After=network.target
+
+        [Service]
+        ExecStart=/usr/sbin/lldpd -d
```
Frostman, line 39: Hm, I don't think it's available on the flatcar

Frostman (changes requested):

pkg/hhfab/vlab_butane.tmpl.yaml
```diff
@@ -55,3 +55,29 @@ storage:
           TOOLBOX_DOCKER_IMAGE=ghcr.io/githedgehog/toolbox
           TOOLBOX_DOCKER_TAG=latest
           TOOLBOX_USER=root
+
+    - path: /etc/systemd/network/enp2s1.network
```
Frostman, line 59: probably something like 

```
[Match]
Name=en*

[Network]
DHCP=yes
LLDP=yes
EmitLLDP=yes
```

in `/etc/systemd/network/99-wildcard.network` would be more generic and will make sure DHCP is always enabled as well (number prefix is generally recommended for drop-ins)

pau-hedgehog: Had tried wildcards without success. But I've tried this and works like a charm. Thanks!

pau-hedgehog: I was facing issues with this approach. The fact that servers are sending LLDP on both the physical and the vlan interfaces is not welcomed by the sonic counterpart

Frostman (approved):

PR #300, opened 2025-01-09
Vlab show tech

pau-hedgehog:
Gathers outputs from the different VMs/SWs in the VLAB to allow forensic analysis when a vlab up ends/fails

Fixes https://github.com/githedgehog/fabricator/issues/335


Frostman:
@pau-hedgehog looks like it's not collecting data when running the upgrade scenario

pau-hedgehog:
> @pau-hedgehog looks like it's not collecting data when running the upgrade scenario

For each upgrade job there should be 2 artifacts, `old` and `cur(rent)`. But:
```
      - name: Initial VLAB with old version
        env:
          HHFAB_REG_REPO: 127.0.0.1:30000
        run: |
          mkdir old
          curl -fsSL https://i.hhdev.io/hhfab | USE_SUDO=false INSTALL_DIR=./old VERSION="${{ matrix.fromversion }}" bash

          old/hhfab init -v --dev -m ${{ matrix.fabricmode }} --include-onie=true
          old/hhfab vlab gen -v
          old/hhfab vlab up -v --ready setup-vpcs --ready test-connectivity --ready exit ${{ matrix.fromflags }}

      - name: Upload show-tech artifacts
        uses: actions/upload-artifact@v4
        if: ${{ always() }}
        with:
          name: show-tech-upgrade-old-${{ matrix.fabricmode }}-${{ matrix.fromversion }}-${{ matrix.fromflags == '--usb' && 'usb' || matrix.fromflags == '-m=iso' && 'iso' }}
          path: show-tech-output
```

But it seems `VLABShowTech` is not being executed prior to exit:
```
                                case OnReadyExit:
                                        if err := c.VLABShowTech(ctx, vlab); err != nil {
                                                slog.Warn("Failed to collect show-tech diagnostics", "err", err)

                                                return fmt.Errorf("getting show-tech: %w", err)
                                        }
```

Frostman (approved):

PR #338, opened 2025-01-22
Automated test plan

edipascale:


Frostman:
@edipascale could you please make a separate PR with the things refactoring in the testing.go so we can get it merged quickly and you'll not have to rebase

.github/workflows/ci.yaml
```diff
@@ -424,10 +426,28 @@ jobs:
...
-      - name: hhfab init and vlab up
+      - name: hhfab release test
+        env:
+          HHFAB_REG_REPO: 127.0.0.1:30000
+          HHFAB_VLAB_COLLECT: true
+        if: ${{ env.RunReleaseTest }}
+        run: |
+          source "./lab-ci/envs/$KUBE_NODE/source.sh"
+          bin/hhfab init -v --dev --include-onie=${{ matrix.includeonie }} -w "./lab-ci/envs/$KUBE_NODE/wiring.yaml"
+          bin/hhfab vlab up -v --ready switch-reinstall --ready wait --ready release-test --ready exit --mode=${{ matrix.buildmode }}
```
Frostman, line 437: We should use `--ready inspect` instead of `--ready wait` and I think it'll be great to keep the standard test for uniformity of results (`--ready setup-vpcs --ready test-connectivity`) before running the release tests

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	tcOpts           TestConnectivityOpts
+	extName          string
+	hhfabBin         string
+}
+
+// prepare for a test: wipe the fabric and then create the VPCs according to the
+// options in the test context
+func (testCtx *VPCPeeringTestCtx) setupTest(ctx context.Context) error {
+	if err := hhfctl.VPCWipeWithClient(ctx, testCtx.kube); err != nil {
+		return errors.Wrap(err, "WipeBetweenTests")
```
Frostman, line 49: When wrapping error please provide context in a human readable message, assuming migration to a new errors smth like:

`return fmt.Errorf("wiping vpcs between tests: %w", err)`

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	}
+	if err := DoVLABSetupVPCs(ctx, testCtx.workDir, testCtx.cacheDir, testCtx.opts); err != nil {
+		return errors.Wrap(err, "DoVLABSetupVPCs")
```
Frostman, line 52: Same here, I'll not explicitly mark other occurencies

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+			},
+		},
+	}
+}
+
+// populate the vpcPeerings map with all possible VPC peering combinations
+func populateFullMeshVpcPeerings(ctx context.Context, kube client.Client, vpcPeerings map[string]*vpcapi.VPCPeeringSpec) error {
+	vpcs := &vpcapi.VPCList{}
+	if err := kube.List(ctx, vpcs); err != nil {
+		return errors.Wrapf(err, "kube.List")
```
Frostman, line 106: For such cases as well, we want to get little bit more context, e.g. "listing vpcs: %w"

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	)
+	cmd.Dir = workDir
+	out, err := cmd.CombinedOutput()
+	pingOK := err == nil && strings.Contains(string(out), "0% packet loss")
+	if expectSuccess && !pingOK {
+		slog.Error("Ping failed, expected success", "source", nodeName, "target", ip, "error", err)
+		slog.Debug("Output of ping", "output", string(out))
+
+		return errors.New("ping failed, expected success")
+	} else if !expectSuccess && pingOK {
```
Frostman, line 253: What if we expect ping to fail (expectSuccess is false), but only some of packets are lost (pingOK MAY be false too? In theory ping should return non-zero code which will cause cmd to fail with error, let's confirm it) - in that case we'll just return nil like there is no issues.

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+
+// disable agent, shutdown port, test connectivity, enable agent, set port up
+func shutDownPortAndTest(ctx context.Context, testCtx *VPCPeeringTestCtx, deviceName string, nosPortName string) error {
+	// disable agent
+	if err := changeAgentStatus(testCtx.hhfabBin, testCtx.workDir, deviceName, false); err != nil {
+		return errors.Wrap(err, "failed to disable agent")
+	}
+	defer func() {
+		if err := changeAgentStatus(testCtx.hhfabBin, testCtx.workDir, deviceName, true); err != nil {
+			slog.Error("Failed to enable agent", "error", err)
```
Frostman, line 411: As a TODO, no need to fix now: It feels like we may want to retry it as if we aren't able to bring agent back future tests will fail and/or we'll keep env in a broken state

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+		}
+	}()
+
+	// set port down
+	if err := changeSwitchPortStatus(testCtx.hhfabBin, testCtx.workDir, deviceName, nosPortName, false); err != nil {
+		return errors.Wrap(err, "failed to set port down")
+	}
+	defer func() {
+		if err := changeSwitchPortStatus(testCtx.hhfabBin, testCtx.workDir, deviceName, nosPortName, true); err != nil {
+			slog.Error("Failed to set port up", "error", err)
```
Frostman, line 421: Not so bad in this case as agent should enforce it eventually

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+			for _, link := range conn.Spec.Fabric.Links {
+				spinePort := link.Spine.LocalPortName()
+				nosPortName, ok := portMap[spinePort]
+				if !ok {
+					return false, errors.Errorf("Port %s not found in switch profile %s for switch %s", spinePort, profile.Name, spine.Name)
+				}
+				if err := changeSwitchPortStatus(testCtx.hhfabBin, testCtx.workDir, spine.Name, nosPortName, false); err != nil {
+					return false, errors.Wrap(err, "failed to set port down")
+				}
+				// XXX: do we need to set ports back up? restarting the agent should eventually take care of that
```
Frostman, line 643: Should it be converted to a TODO or removed?

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+		sub.Restricted = pointer.To(false)
+	}
+	_, err = CreateOrUpdateVpc(ctx, testCtx.kube, vpc)
+	if err != nil {
+		return false, errors.Wrap(err, "CreateOrUpdateVpc")
+	}
+	_, err = CreateOrUpdateVpc(ctx, testCtx.kube, vpc2)
+	if err != nil {
+		return false, errors.Wrap(err, "CreateOrUpdateVpc")
+	}
```
Frostman, line 751: Should we test connectivity again afer removing restrictions?

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	slog.Debug("Removing all restrictions")
+	vpc.Spec.Permit = make([][]string, 0)
+	for _, sub := range vpc.Spec.Subnets {
+		sub.Isolated = pointer.To(false)
+		sub.Restricted = pointer.To(false)
+	}
+	_, err = CreateOrUpdateVpc(ctx, testCtx.kube, vpc)
+	if err != nil {
+		return false, errors.Wrap(err, "CreateOrUpdateVpc")
+	}
```
Frostman, line 862: Same here - would be good to test connectivity IMO

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+		newVpcAtt := &vpcapi.VPCAttachment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      vpcAtt.Name,
+				Namespace: vpcAtt.Namespace,
+			},
+			Spec: vpcAtt.Spec,
+		}
+		slog.Debug("Creating VPCAttachment", "attachment", newVpcAtt.Name)
+		if err := testCtx.kube.Create(ctx, newVpcAtt); err != nil {
+			slog.Error("Error creating VPCAttachment", "error", err)
```
Frostman, line 931: As a TODO, no need to fix now: Failing this defer would keep fabric in the "broken" state

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	slog.Debug("Adding dummy inteface with address 10.199.0.100/32 to the server", "server", server)
+	if err := execNodeCmd(testCtx.hhfabBin, testCtx.workDir, server, "sudo ip link add dummy0 type dummy"); err != nil {
+		return false, errors.Wrap(err, "execNodeCmd")
+	}
+	if err := execNodeCmd(testCtx.hhfabBin, testCtx.workDir, server, "sudo ip addr add 10.199.0.100/32 dev dummy0"); err != nil {
+		return false, errors.Wrap(err, "execNodeCmd")
+	}
+	defer func() {
+		slog.Debug("Removing address and default route from en2ps1 on the server", "server", server)
+		_ = execNodeCmd(testCtx.hhfabBin, testCtx.workDir, server, "sudo ip addr del 172.30.50.1/24 dev enp2s1")
```
Frostman, line 1035: ignoring the errors from cmd execs will completely hide from is we've been unable to conf ips or cleanup

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+		time.Sleep(5 * time.Second)
+		if err := WaitSwitchesReady(ctx, testCtx.kube, 0, 5*time.Minute); err != nil {
+			slog.Error("Error waiting for switches to be ready", "error", err)
+
+			return
+		}
+		if err := execNodeCmd(testCtx.hhfabBin, testCtx.workDir, "server-1", "/opt/bin/hhnet cleanup"); err != nil {
+			slog.Error("Error cleaning up server-1", "error", err)
+		}
+		// FIXME: ideally this would be derived rather than hardcoded (extract the code from testing.go)
```
Frostman, line 1113: Convert to TODO? I think it's okay for this PR

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	if err := WaitSwitchesReady(ctx, testCtx.kube, 0, 5*time.Minute); err != nil {
+		return false, errors.Wrap(err, "WaitSwitchesReady")
+	}
+
+	// Configure network on server-1
+	slog.Debug("Configuring network on server-1")
+	if err := execNodeCmd(testCtx.hhfabBin, testCtx.workDir, "server-1", "/opt/bin/hhnet cleanup"); err != nil {
+		return false, errors.Wrap(err, "execNodeCmd")
+	}
+	// FIXME: ideally this would be derived rather than hardcoded (extract the code from testing.go)
```
Frostman, line 1130: Convert to TODO? I think it's okay for this PR

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+			F:    testCtx.vpcPeeringsSergeisSpecialTest,
+		},
+	}
+	suite.Tests = len(suite.TestCases)
+
+	return suite
+}
+
+func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOtps ReleaseTestOpts) error {
+	// FIXME: make this configurable
```
Frostman, line 1440: Convert to TODO? I think it's okay for this PR

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+}
+
+type Skipped struct {
+	XMLName xml.Name `xml:"skipped"`
+	Message string   `xml:"message,attr,omitempty"`
+}
+
+func printSuiteResults(ts *JUnitTestSuite) {
+	var numFailed, numSkipped, numPassed int
+	slog.Info(fmt.Sprintf("Results for %s:", ts.Name))
```
Frostman, line 1214: slog is supposed to be used with args smth like slog.Info("Results", "testSuite", ts.Name)

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	for _, test := range ts.TestCases {
+		if test.Skipped != nil {
+			slog.Info(fmt.Sprintf("SKIP %s", test.Name), "skip-message", test.Skipped.Message)
```
Frostman, line 1217: Same, and it'll be more clear (and parseable) - `slog.Info("SKIP", "test", test.Name, "reason", test.Skipped.Message)

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+			numSkipped++
+		} else if test.Failure != nil {
+			slog.Warn(fmt.Sprintf("FAIL %s", test.Name), "error", strings.Split(test.Failure.Message, "\n")[0])
+			numFailed++
+		} else {
+			slog.Info(fmt.Sprintf("PASS %s", test.Name))
```
Frostman, line 1223: Here and later for `slog.Info(fmt.Sprintf`

Frostman (changes requested):
Great job! I wasn't digging much into each test case, mainly looking for potential issues, unclear behavior, and general Go code issues.

Few notes for fixing in future PRs:
- I have a few questions about making sure that tests that are suitable for the VLAB will run
- failed cleanups in defer may leave fabric in a semi-broken state

pkg/hhfab/release.go
```diff
@@ -0,0 +1,1499 @@
...
+	// Create new connection with static external
+	staticExtConn := &wiringapi.Connection{}
+	staticExtConn.Name = fmt.Sprintf("release-test--static-external--%s", switchName)
+	staticExtConn.Namespace = "default"
+	staticExtConn.Spec.StaticExternal = &wiringapi.ConnStaticExternal{
+		WithinVPC: "vpc-01",
+		Link: wiringapi.ConnStaticExternalLink{
+			Switch: wiringapi.ConnStaticExternalLinkSwitch{
+				BasePortName: wiringapi.NewBasePortName(switchPortName),
+				IP:           "172.30.50.5/24",
```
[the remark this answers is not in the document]
edipascale: changed to 172.31.255.X

edipascale:
@Frostman I think I've addressed all of your comments. A couple of standing issues:
- a run now takes ~2h5m in env-1. I tried cutting corners but short of disabling iperfs (which we agreed we cannot afford to do) the only other option I can think of is to skip entire tests. There is an extended flag which is set to false by default, but currently it only disables curls and some extra steps in a few tests. Infra is already there to skip entire tests based on that flag, if desired.
- vlab release-tests in the ci have been enabled using the same label trick, but based on many runs on my local machine I had to disable several tests, i.e. all failover ones, the static external and all of those with "restrictions". Then there are those that won't run because of the lack of externals, and the final result is that there's only a couple of tests left.
- even on real envs, tests seem to be quite brittle. I've had a few runs with no issues, but many more with one/two random failures, typically on a ping check, pretty much on any of the tests with no discernible pattern. I've also seen one failure with a slightly below target iperf speed (8490 vs 8500). Unless I missed something, we are already waiting for switches to be ready with an "applied for" timer of at least 30 seconds in all instances where I've observed failures, but it doesn't seem to be enough. I could try to increase the timer further, but that would make the runs even longer, and regardless I'm not sure where we should set the acceptance bar or how long is reasonable to wait for

Feel free to have a look and/or let me know if you want to discuss this over a call!

edipascale:
one additional data point: after bumping the apppliedFor timer to 1m everywhere, I still got a failure in env-1:
```
10:37:54 ERR FAIL test="Full Mesh All Externals" error="testing connectivity: testing connectivity: testing connectivity from \"server-1\" to \"server-3\": checking ping from server-1 to server-3: should be reachable but ping failed with output: PING 10.10.3.2 (10.10.3.2) 56(84) bytes of data."
[...]
10:37:54 INF All tests completed duration=2h12m32.561923603s
10:37:54 ERR release-test: some tests failed: singleVpc=0, multiVpc=0, basic=1
```
the length of the tests did not increase by too much, I'll try 2 minutes next

edipascale:
> the length of the tests did not increase by too much, I'll try 2 minutes next

First run with a 2 minutes timer succeeded, but took considerably longer:
```
16:16:15 INF All tests completed duration=3h0m32.49154262s
```

I'm going to run a few more to see if it's consistent

edipascale:
> First run with a 2 minutes timer succeeded, but took considerably longer:
> 
> ```
> 16:16:15 INF All tests completed duration=3h0m32.49154262s
> ```
> 
> I'm going to run a few more to see if it's consistent

it isn't :(
```
19:25:04 INF All tests completed duration=3h1m39.01107962s
19:25:04 ERR release-test: some tests failed: singleVpc=1, multiVpc=0, basic=1
```

Frostman (approved):

Frostman (approved):

Issue #310, opened 2025-01-13
VLAB fails to SSH to server or client test endpoint


pau-hedgehog:
Observed SSH connection to server failure in:
https://github.com/githedgehog/fabricator/actions/runs/12743481149/job/35513736851

```
09:27:19 DBG Ping result from=server-06 to=server-01 expected=false ok=false fail=true err="Process exited with status 1" out="PING 10.0.1.2 (10.0.1.2) 56(84) bytes of data.\n\n--- 10.0.1.2 ping statistics ---\n5 packets transmitted, 0 received, 100% packet loss, time 4121ms"
09:27:21 DBG IPerf3 result from=server-03 to=server-04 sendSpeed="4.16 Mbps" receiveSpeed="3.94 Mbps" sent="2.60 MB" received="2.47 MB"
09:27:21 DBG Running iperf from=server-06 to=server-05
09:27:27 DBG IPerf3 result from=server-06 to=server-05 sendSpeed="27.16 Mbps" receiveSpeed="26.66 Mbps" sent="16.97 MB" received="16.79 MB"
09:27:27 DBG Running iperf from=server-10 to=server-09
09:27:34 DBG IPerf3 result from=server-10 to=server-09 sendSpeed="8.05 Mbps" receiveSpeed="7.73 Mbps" sent="5.03 MB" received="4.86 MB"
09:27:34 DBG Running iperf from=server-08 to=server-07
09:27:40 DBG IPerf3 result from=server-08 to=server-07 sendSpeed="2.71 Mbps" receiveSpeed="2.45 Mbps" sent="1.70 MB" received="1.54 MB"
09:27:40 DBG Running iperf from=server-04 to=server-03
09:27:46 DBG IPerf3 result from=server-04 to=server-03 sendSpeed="4.96 Mbps" receiveSpeed="4.69 Mbps" sent="3.10 MB" received="2.94 MB"
09:27:46 DBG Running iperf from=server-05 to=server-06
09:27:53 DBG IPerf3 result from=server-05 to=server-06 sendSpeed="11.22 Mbps" receiveSpeed="10.91 Mbps" sent="7.01 MB" received="6.87 MB"
09:27:53 DBG Running iperf from=server-07 to=server-08
09:27:59 DBG IPerf3 result from=server-07 to=server-08 sendSpeed="2.72 Mbps" receiveSpeed="2.52 Mbps" sent="1.70 MB" received="1.58 MB"
[...]

PR #342, opened 2025-01-24
Add retries to SSH iperf on server

pau-hedgehog:
Extend show tech logs gathered on server VMs

Fixes https://github.com/githedgehog/fabricator/issues/310

### DIFF
--- a/pkg/hhfab/testing.go
+++ b/pkg/hhfab/testing.go
@@ -1261,74 +1261,84 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
 	if opts.IPerfsSeconds <= 0 || !expected {
 		return nil
 	}
+	maxRetries := 3
+	var lastErr error
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		if err := iperfs.Acquire(ctx, 1); err != nil {
+			return fmt.Errorf("acquiring iperf semaphore: %w", err)
+		}
 
-	if err := iperfs.Acquire(ctx, 1); err != nil {
-		return fmt.Errorf("acquiring iperf semaphore: %w", err)
-	}
-	defer iperfs.Release(1)
+		err := func() error {
+			defer iperfs.Release(1)
+			attemptCtx, cancel := context.WithTimeout(ctx, time.Duration(opts.IPerfsSeconds+30)*time.Second)
+			defer cancel()
 
-	ctx, cancel := context.WithTimeout(ctx, time.Duration(opts.IPerfsSeconds+30)*time.Second)
-	defer cancel()
+			slog.Debug("Running iperf", "from", from, "to", to, "attempt", attempt)
+			g, attemptCtx := errgroup.WithContext(attemptCtx)
 
-	slog.Debug("Running iperf", "from", from, "to", to)
-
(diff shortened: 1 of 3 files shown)


pkg/hhfab/testing.go
```diff
@@ -1201,12 +1201,20 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
 	g, ctx := errgroup.WithContext(ctx)
 
 	g.Go(func() error {
-		out, err := toSSH.RunContext(ctx, fmt.Sprintf("toolbox -q timeout -v %d iperf3 -s -1", opts.IPerfsSeconds+25))
-		if err != nil {
-			return fmt.Errorf("running iperf server: %w: %s", err, string(out))
+		maxRetries := 3
```
Frostman, line 1204: the context.WithTimeout is built around iperfsseconds so there is no "time" to retry and if we're restarting the server we most probably need to restart the client as well. It maybe makes sense to retry the whole checkIPerf function

pkg/hhfab/testing.go
```diff
@@ -1188,65 +1188,87 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
...
-	defer cancel()
+	maxRetries := 3
+	var lastErr error
 
-	slog.Debug("Running iperf", "from", from, "to", to)
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		if err := iperfs.Acquire(ctx, 1); err != nil {
+			return fmt.Errorf("acquiring iperf semaphore: %w", err)
+		}
+		defer iperfs.Release(1)
```
Frostman, line 1198: Defer in the loop will be called when the parent function containing the loop finished, not when one iteration of the loop is finished so it'll stuck forever after a first attempt

Frostman (changes requested):
Please fix defer in the loop usage, one option is to wrap the loop body into the inline func and call it immediately:

```
for {
  func() { ... }()
}

File: pkg/hhfab/testing.go [hunk 3]
```diff
@@ -1261,66 +1261,84 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
...
+			attemptCtx, cancel := context.WithTimeout(ctx, time.Duration(opts.IPerfsSeconds+30)*time.Second)
+			defer cancel()
 
-	ctx, cancel := context.WithTimeout(ctx, time.Duration(opts.IPerfsSeconds+30)*time.Second)
-	defer cancel()
+			slog.Debug("Running iperf", "from", from, "to", to, "attempt", attempt)
+			g, attemptCtx := errgroup.WithContext(attemptCtx)
 
-	slog.Debug("Running iperf", "from", from, "to", to)
+			g.Go(func() error {
```
edipascale, line 1279: There is something I don't understand... If I read the docs on golang's errgroup correctly, each of these functions will be started immediately in a separate goroutine (as we are not setting any limit on the errgroup), but the operations should actually be sequential, or at least partly so - i.e. starting the server and checking for it in the client could be done in parallel, although I'm not quite sure what the benefits are, but you definitely do not want to start the actual client until the server is ready, especially since that operation does not have a retry. Did I misinterpret something here?

PR #347, opened 2025-01-27
hhfab: splits and helpers for testing

edipascale:
split some of the CLI functions and expose some helpers so that a test program can call/build upon the existing code instead of having to run the executable.


edipascale:
@Frostman I created this separate PR with the base changes for the automated tests as requested, can you review it when you get the chance? thanks!

Frostman (approved):

Issue #333, opened 2025-01-21
vlab upgrade control node USB auto installation failure


pau-hedgehog:
Observed in https://github.com/githedgehog/fabricator/actions/runs/12876382887/job/35899546444

```
22:35:47 DBG SSH is ready vm=control-1 type=control
22:35:47 DBG Waiting for control node to be auto installed (via USB) vm=control-1 type=control
23:03:12 WRN Failed to post-process VM vm=control-1 type=control err="cancelled: context deadline exceeded"
23:03:27 DBG Force exit with code 2 err="context canceled"
```

It seems the control-1 VM didn't boot properly. We should enhance show-tech to capture information of the host/runner, maybe

pau-hedgehog:
Another hit: https://github.com/githedgehog/fabricator/actions/runs/12894243152/job/35978390575

pau-hedgehog:
Another hit: https://github.com/githedgehog/fabricator/actions/runs/12983690706/job/36222415114

Will have to extend show-tech to run in this case and gather enough diagnostic information

edipascale:
one more hit: https://github.com/githedgehog/fabricator/actions/runs/13542872463/job/37848156503

pau-hedgehog:
After the new runners change I'm not seeing this anymore

PR #349, opened 2025-01-27
Trigger show-tech in case of VM postProcess timeout

pau-hedgehog:
We are facing several CI failure conditions where show-tech is not executed and thus we cannot diagnose the root cause

Related to https://github.com/githedgehog/fabricator/issues/333

### DIFF
--- a/pkg/hhfab/vlabrunner.go
+++ b/pkg/hhfab/vlabrunner.go
@@ -400,24 +400,25 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 		if vm.Type == VMTypeServer || vm.Type == VMTypeControl {
 			postProcesses.Add(1)
 			group.Go(func() error {
+				defer postProcesses.Done()
+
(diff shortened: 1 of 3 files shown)


pau-hedgehog:
Maybe we should also try to capture show-tech when other on-ready commands fail:

https://github.com/githedgehog/fabricator/actions/runs/13070506705/job/36471259713

pkg/hhfab/vlabrunner.go
```diff
@@ -400,25 +400,26 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 		if vm.Type == VMTypeServer || vm.Type == VMTypeControl {
 			postProcesses.Add(1)
 			group.Go(func() error {
+				defer postProcesses.Done()
```
Frostman, line 403: I think having defer ...Done() will break the logic of post actions, I need to dig deeper to understand if it's a problem

pau-hedgehog:
I think we need another approach to manage the VLAB life-cycle and show-tech gathering

PR #354, opened 2025-01-29
vlab and hlab stability troubleshooting and improvements

### LINKED ISSUE #292: VLAB random setup-vpcs errors when configuring server VMs
There are known issues with the `hhnet` script: https://github.com/githedgehog/fabricator/blob/master/pkg/hhfab/hhnet.sh

```
23:54:56 WRN Error running on-ready commands err="setting up VPCs: configuring servers: configuring server \"server-08\": running hhnet \"bond 1008 enp2s1 enp2s2\": Process exited with status 1: out: Failed to get IP address for bond0.1008\n"
```
But as it's hitting the CI from time to time it needs to be addressed to make the CI more reliable and avoid having to retry the job:
https://github.com/githedgehog/fabricator/actions/runs/12530586269/job/34947281228

pau-hedgehog:
Migrate to networkctl for stability

Fixes #292

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -255,6 +255,15 @@ jobs:
         run: |
           cat .zot/log
 
+      - name: Collect VM Serial Logs
+        if: ${{ always() }}
+        run: |
+          for vm in bin/hhfab/vlab/vms/*; do
+            if [ -f "$vm/serial.log" ]; then
+              cp "$vm/serial.log" "show-tech-output/$(basename $vm)-serial.log"
+            fi
+          done
+
       - name: Upload show-tech artifacts
         uses: actions/upload-artifact@v4
         if: ${{ always() }}
@@ -320,6 +329,15 @@ jobs:
           old/hhfab vlab gen -v
           old/hhfab vlab up -v --ready setup-vpcs --ready test-connectivity --ready exit ${{ matrix.fromflags }}
 
+      - name: Collect VM Serial Logs
+        if: ${{ always() }}
+        run: |
+          for vm in bin/hhfab/vlab/vms/*; do
+            if [ -f "$vm/serial.log" ]; then
+              cp "$vm/serial.log" "show-tech-output/$(basename $vm)-serial.log"
+            fi
+          done
+
       - name: Upload show-tech artifacts
         uses: actions/upload-artifact@v4
         if: ${{ always() }}
@@ -337,6 +355,15 @@ jobs:
           HHFAB_VLAB_COLLECT: true
         run: |
           bin/hhfab vlab up -v --ready setup-vpcs --ready inspect --ready test-connectivity --ready exit --upgrade
+
+      - name: Collect VM Serial Logs
+        if: ${{ always() }}
+        run: |
+          for vm in bin/hhfab/vlab/vms/*; do
+            if [ -f "$vm/serial.log" ]; then
+              cp "$vm/serial.log" "show-tech-output/$(basename $vm)-serial.log"
+            fi
+          done
 
       - name: Upload show-tech artifacts
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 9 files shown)


Frostman:
@pau-hedgehog, I'm not arguing about whether it's more stable, but what kind of change do you see with that change?

pkg/hhfab/hhnet.sh
```diff
@@ -2,45 +2,137 @@
...
+Name=$iface
+[Network]
+Bond=$bond_name
+[Link]
+MTUBytes=9036
+EOF
     done
 
-    sudo ip l s "$bond_name" up
+    sudo systemctl restart systemd-networkd 2> /dev/null || true
```
Frostman, line 61: In theory, `networkctl reload` should reconfigure the network, but restart is probably more reliable

pkg/hhfab/hhnet.sh
```diff
@@ -2,45 +2,137 @@
...
 
-
 set -e
 
+NETWORKD_PATH="/etc/systemd/network"
+
 function cleanup() {
     for i in {0..3}; do
-        sudo ip l d "bond$i" 2> /dev/null || true
+        sudo networkctl down "bond$i" 2> /dev/null || true
```
Frostman, line 11: Do we even need to down bonds/interfaces/vlans? If we're removing the dropins and the restarting networkd it should do all the cleanup

Frostman:
@pau-hedgehog no, using networkd completely makes sense to me, I think it's an artifact from the very early stage when it was probably not even a flatcar but ubuntu on the test servers

Frostman:
I think it may break the LLDP as only a first matching dropin will be used, so, you maybe need to enable LLDP on the slave interfaces, not sure

pkg/hhfab/hhnet.sh
pau-hedgehog: I'm trying with this

pkg/hhfab/hhnet.sh
pau-hedgehog: I think the cleanup will be more idempotent this way

pau-hedgehog:
> I think it may break the LLDP as only a first matching dropin will be used, so, you maybe need to enable LLDP on the slave interfaces, not sure

good point. It's not working. I'll make sure the inspect passes

.github/workflows/ci.yaml
```diff
@@ -245,14 +246,25 @@ jobs:
           HHFAB_VLAB_COLLECT: true
         run: |
           bin/hhfab init -v --dev -m ${{ matrix.fabricmode }} --include-onie=${{ matrix.includeonie }}
+          bin/hhfab versions
           bin/hhfab vlab gen -v
-          bin/hhfab vlab up -v --ready inspect --ready setup-vpcs --ready test-connectivity --ready exit --mode=${{ matrix.buildmode }}
+          bin/hhfab vlab up -v --ready setup-vpcs --ready inspect --ready test-connectivity --ready exit --mode=${{ matrix.buildmode }}
```
Frostman, line 251: Why set up VPCs first? We should see LLDP from the servers even before configuring the ports

pkg/hhfab/cmdvlab.go
```diff
@@ -217,6 +217,7 @@ type SwitchReinstallOpts struct {
 	PDUUsername    string              // (hard-reset mode only)
 	PDUPassword    string              // (hard-reset mode only)
 	WaitReady      bool                // Wait for the switch to be ready
+	Verbose        bool                // To relay the verbose flag
```
Frostman, line 220: You don't need to pass Verbose, just use `slog.Default().Enabled(ctx, slog.LevelDebug)`

.github/workflows/ci.yaml
pau-hedgehog: This was temporary to reduce the pipeline failure rate and be able to keep troubleshooting other failures

pkg/hhfab/cmdvlab.go
Frostman: there os already -v flag on all commands and it sets logging to debug

PR #356, opened 2025-01-31
Hhnet mtu 9036

pau-hedgehog:
Related to https://github.com/githedgehog/internal/issues/83


pau-hedgehog:
@Frostman and I decided it's ok to keep the current MTU as the main purpose of the iperf tests is to rule out traffic going through the CPU

PR #369, opened 2025-02-04
Added flatcar upgrade bin to the installer artifacts

cesargithedgehog:


Frostman:
- I've pulled in your fab/README.md commit - it should be part of this PR
- We don't allow Merge commits in the history
- DCO should be passing https://github.com/githedgehog/fabricator/pull/369/checks?check_run_id=36747883694

PR #392, opened 2025-02-12
In-place airgap Flatcar upgrade

cesargithedgehog:


pkg/fab/README.md
```diff
@@ -16,9 +16,12 @@ mv flatcar_production_qemu_image.img flatcar.img
 mv flatcar_production_qemu_uefi_efi_code.fd flatcar_efi_code.fd
```
Frostman, line 16: Please bump the versing in the beginning of that snippet as well (`export FLATCAR_VERSION`)

pkg/fab/comp/flatcar/flatcarupdate.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2025 Hedgehog
```
Frostman, line 1: Please use the existing file (flatcar.go)

pkg/fab/comp/flatcar/flatcarupdate.go
```diff
@@ -0,0 +1,18 @@
...
+
+package flatcar
+
+import (
+	fabapi "go.githedgehog.com/fabricator/api/fabricator/v1beta1"
+	"go.githedgehog.com/fabricator/api/meta"
+)
+
+const (
+	Ref     = "fabricator/flatcar-update"
```
Frostman, line 12: Should be `UpdateRef`

pkg/fab/comp/flatcar/flatcarupdate.go
```diff
@@ -0,0 +1,18 @@
...
+	BinName = "flatcar_production_update.gz"
```
Frostman, line 13: Same-ish: `UpdateBinName`

pkg/fab/recipe/control_upgrade.go
```diff
@@ -179,11 +179,55 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+
 	slog.Info("Control node upgrade complete")
 
 	return nil
 }
 
+func (c *ControlUpgrade) upgradeFlatcar(ctx context.Context) error {
+	flatcarVersion := string(c.Fab.Status.Versions.Fabricator.Flatcar)
+
+	cmd := exec.CommandContext(ctx, "grep", "^VERSION=", "/etc/os-release")
```
Frostman, line 194: Could you please read file instead of running commands (`os.ReadFile`)

pkg/fab/recipe/control_upgrade.go
```diff
@@ -179,11 +179,55 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+		return fmt.Errorf("Flatcar Version: %w", err)
+	} else if ret := strings.Compare(string(out[:]), "VERSION="+flatcarVersion[1:]+"\n"); ret == 0 {
+		slog.Info("System already updated running Flatcar", "version", flatcarVersion)
+
+		return nil
+	}
+
+	slog.Info("Upgrading Flatcar to", "version", flatcarVersion)
+
+	ctx1, cancel := context.WithTimeout(ctx, time.Second*300)
```
Frostman, line 206: It's probably better to move that to the beginning of the function and there is no need to name ctx1, you can shadow the ctx var. Additionally, 5*time.Minutes would be more readable.

pkg/fab/recipe/control_upgrade.go
```diff
@@ -179,11 +179,55 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+	cmd1.Stderr = logutil.NewSink(ctx1, slog.Debug, "update-flatcar: ")
+
+	if err := cmd1.Run(); err != nil {
+		return fmt.Errorf("upgrading Flatcar: %w", err)
+	}
+
+	slog.Info("Flatcar upgrade completed")
+	slog.Info("Rebooting Control Node")
+
+	cmd2 := exec.CommandContext(ctx, "reboot")
```
Frostman, line 220: no need for cmd1/cmd2, you can use a single cmd var

pkg/fab/recipe/control_upgrade.go
```diff
@@ -179,11 +179,55 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+	cmd2.Stdout = logutil.NewSink(ctx, slog.Debug, "reboot: ")
+	cmd2.Stderr = logutil.NewSink(ctx, slog.Debug, "reboot: ")
+
+	if err := cmd2.Run(); err != nil {
+		return fmt.Errorf("Rebooting: %w", err)
```
Frostman, line 225: when wrapping errors message should start with lower case

Frostman (changes requested):

pkg/fab/recipe/control_upgrade.go
```diff
@@ -179,11 +179,55 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+
+		return nil
+	}
+
+	slog.Info("Upgrading Flatcar to", "version", flatcarVersion)
+
+	ctx1, cancel := context.WithTimeout(ctx, time.Second*300)
+	defer cancel()
+
+	cmd1 := exec.CommandContext(ctx1, "flatcar-update", "--to-version", flatcarVersion, "--to-payload", "/opt/hedgehog/install/control-1-install/flatcar_production_update.gz")
```
Frostman, line 209: path shouldn't be hardcoded, instead you can use a relative path and you should use `flatcar.UpdateBinName` const

Frostman:
Additionally, please cleanup commit history ("fix"/"fix2"), it's totally okay to have multiple commits just with proper summaries

pkg/fab/recipe/control_upgrade.go
Frostman: Additionally, let's please have it a bit more safe:
- prompt for user to accept reboot if it's a terminal (you can use `isatty.IsTerminal` to check it) 
- add `--yes` flag to the recipe cli to auto-accept
- fail if no flag and it's not a terminal

You'll need to make sure that vlab is calling it with --yes flag for the upgrade automation (vlabrunner.go)

pkg/fab/recipe/control_upgrade.go
```diff
@@ -180,11 +185,118 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+		} else if response == "n" || response == "no" {
+			return false
+		}
+	}
+}
+
+func (c *ControlUpgrade) upgradeFlatcar(ctx context.Context) error {
+	slog.Info("Upgrading Flatcar")
+	filename := "/etc/os-release"
+	autoReboot := false
```
Frostman, line 221: I think there is no need to separate flatcar upgrade and reboot. We need to apply upgrade after the version check anyways, yes flag (or confirmation) is only needed for a reboot

pkg/fab/recipe/control_upgrade.go
```diff
@@ -180,11 +185,118 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+
+		if response == "y" || response == "yes" {
+			return true
+		} else if response == "n" || response == "no" {
+			return false
+		}
+	}
+}
+
+func (c *ControlUpgrade) upgradeFlatcar(ctx context.Context) error {
```
Frostman, line 218: Just putting a summary comment here. I think this function should have following logical structure:

- check for a flatcar version, if it's as desired just return without error as there is nothing to do
- run flatcar-upgrade
- if yes flag is provided or it's a terminal and prompt for confirmation is positive then reboot
- \\ otherwise just log that reboot is required to finish the flatcar upgrade (no need to return error in that case - it's valid)

Which overall means that --yes and prompt for confirmation just controls if we'll perform reboot or not and will not result in an error

Frostman (changes requested):

Frostman:
@cesargithedgehog did you had a chance to manually test the upgrade from beta-1 - same way as it's done in the CI? Failures of the vlab-upgrade jobs looks legit.

Frostman:
Additionally, it seems to be breaking when using manual mode (-m=manual) which is using prebiult qemu flatcar image with the following error:

```
16:27:05 INF control-upgrade: 00:27:05 DBG K8s node ready with new version version=v1.32.1+k3s1 vm=control-1
16:27:05 INF control-upgrade: 00:27:05 INF Upgrading Flatcar vm=control-1
16:27:05 INF control-upgrade: 00:27:05 INF Upgrading Flatcar to version=v4152.2.0 vm=control-1
16:27:05 INF control-upgrade: 00:27:05 DBG flatcar-update: Error: system requires 'qemu' OEM extension but not passed in --extension vm=control-1
16:27:05 INF control-upgrade: 00:27:05 ERR control upgrade: upgrading Flatcar: upgrading Flatcar: exit status 1 vm=control-1
```

Frostman (changes requested):
Could you please cleanup commit history and rebase to the current master?

- 0644bd553f0bf9ddff4a1a282ff5a10e5ff226e0 seems to be empty?
- f6834e0fe897e3c0db8eb9ed9514625383ca9ae0 includes not your changes

Frostman:
@cesargithedgehog we don't allow merge commits, please rebase your branch instead (and remove 41cff60db8523104f71d765dad48325a52d98598)

.github/workflows/ci.yaml
```diff
@@ -351,6 +351,20 @@ jobs:
...
+          HHFAB_VLAB_COLLECT: true
+        run: |
+          bin/hhfab vlab up -v --ready inspect --ready setup-vpcs --ready test-connectivity --ready exit
+          
+      - name: Upload show-tech artifacts
+        uses: actions/upload-artifact@v4
+        if: ${{ always() }}
+        with:
+          name: show-tech-upgrade-aft-${{ matrix.fabricmode }}-${{ matrix.fromversion }}-${{ matrix.fromflags == '--usb' && 'usb' || matrix.fromflags == '-m=iso' && 'iso' }}
+          path: show-tech-output
```
Frostman, line 366: Please remove trailing whitespaces here

.github/workflows/ci.yaml
```diff
@@ -351,6 +351,20 @@ jobs:
           name: show-tech-upgrade-cur-${{ matrix.fabricmode }}-${{ matrix.fromversion }}-${{ matrix.fromflags == '--usb' && 'usb' || matrix.fromflags == '-m=iso' && 'iso' }}
           path: show-tech-output
 
+      - name: hhfab vlab up after upgrade
+        env:
+          HHFAB_REG_REPO: 127.0.0.1:30000
+          HHFAB_VLAB_COLLECT: true
+        run: |
+          bin/hhfab vlab up -v --ready inspect --ready setup-vpcs --ready test-connectivity --ready exit
```
Frostman, line 359: Try adding `-m=manual` in the end, it should help with the errors I think

Frostman (changes requested):
@cesargithedgehog we don't allow merge commits, please rebase your branch instead (and remove 41cff60db8523104f71d765dad48325a52d98598)

Frostman (approved):

PR #401, opened 2025-02-18
Add vlab up detach and vlab down

pau-hedgehog:
Initial poc of vlab up|down|status to better handle VLAB lifecycle and simplify show-tech gathering


Frostman:
@pau-hedgehog that seems to become stale, should we close it for now? thx

PR #418, opened 2025-02-26
hhfab: add src/dst filters to test-connectivity

edipascale:
so we can use it in demos to generate pings or traffic from specific source-destination pairs

### DIFF
--- a/pkg/hhfab/testing.go
+++ b/pkg/hhfab/testing.go
@@ -874,9 +874,14 @@ type TestConnectivityOpts struct {
 	IPerfsParallel    int64
 	CurlsCount        int
 	CurlsParallel     int64
+	Sources           []string
+	Destinations      []string
 }
 
 func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConnectivityOpts) error {
+	if opts.PingsCount == 0 && opts.IPerfsSeconds == 0 && opts.CurlsCount == 0 {
+		return fmt.Errorf("at least one of pings, iperfs or curls should be enabled")
+	}
 	start := time.Now()
 
 	if opts.PingsParallel <= 0 {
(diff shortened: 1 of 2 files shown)


pkg/hhfab/testing.go
```diff
@@ -947,6 +952,12 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
 
 	serverIDs := map[string]uint64{}
 	for _, server := range servers.Items {
+		// Skip servers not in the list of sources or destinations, if both are specified
+		if len(opts.Sources) > 0 && len(opts.Destinations) > 0 {
```
Frostman, line 956: I think you should split into two conditions so it could be used with only source or only dest specified

Frostman (changes requested):

pkg/hhfab/testing.go
edipascale: you can specify only sources or destinations, but in that case the other group will contain by default all servers, so there is no chance that the current server is not in at least one of the lists and we won't skip it for sure

Frostman: Makes sense

pkg/hhfab/testing.go
```diff
@@ -1060,86 +1071,108 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
...
 
 	g = &errgroup.Group{}
-	for _, serverA := range servers.Items {
-		for _, serverB := range servers.Items {
-			if serverA.Name == serverB.Name {
+	if len(opts.Sources) == 0 {
+		opts.Sources = make([]string, len(serverIDs))
+		// copy keys of serverIDs to opts.Sources
+		i := 0
+		for k := range serverIDs {
```
Frostman, line 1078: Just for the future - you can use `lo.Keys` for it

Frostman (approved):

PR #432, opened 2025-03-05
chore: prohibit using old github.com/pkg/errors

Frostman:


edipascale (approved):
LGTM, although I wonder whether we should just remove err113 from the enabled linters instead of peppering the code with nolint directives? I suppose now we can distinguish between cases where it's relevant and cases where it's not, but I'm not sure that the impact of a miss is worth the hassle. In any case, this is good to go :)

Frostman:
@edipascale , yeah, I think I'm just choosing between peppering the code with suppressing linter and having a huge header with a dozen of errors defined which would never be used outside of it...

PR #434, opened 2025-03-05
ci: fix vlabs and hlabs false positives

pau-hedgehog:
To avoid showing this as pass hlab|vlab:
image


pau-hedgehog:
Raised this: https://github.com/githedgehog/fabricator/issues/442
Closing the PR as it should be addressed in: https://github.com/githedgehog/fabricator/pull/431

PR #440, opened 2025-03-05
ci: make controls unrestricted so ntp upstream works

Frostman:
Seems like we're facing issues caused by switches time getting out of sync over time due to the upstream NTP server not being reacheable from the control node (and so strata is too high).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -445,7 +445,9 @@ jobs:
         run: |
           source "./lab-ci/envs/$KUBE_NODE/source.sh"
           bin/hhfab init -v --dev --include-onie=${{ matrix.includeonie }} -w "./lab-ci/envs/$KUBE_NODE/wiring.yaml"
-          bin/hhfab vlab up -v --ready switch-reinstall --ready inspect --ready setup-vpcs --ready test-connectivity --ready exit --mode=${{ matrix.buildmode }}
+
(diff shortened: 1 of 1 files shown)


edipascale (dismissed):

Frostman:
Ouch, shame on me:


```
Incorrect Usage: flag provided but not defined: -controls-restriced

Did you mean "--controls-restricted"?
```

pau-hedgehog:
The CI failed. I'm re-running it.

I'm wondering if we should always provide time to the switches, even if it drifts

Frostman:
> The CI failed. I'm re-running it.
> 
> I'm wondering if we should always provide time to the switches, even if it drifts

@pau-hedgehog could you please elaborate? I'm not sure I understand what do you mean

pau-hedgehog:
> > The CI failed. I'm re-running it.
> > I'm wondering if we should always provide time to the switches, even if it drifts
> 
> @pau-hedgehog could you please elaborate? I'm not sure I understand what do you mean

I'm talking about this: https://kb.meinbergglobal.com/kb/time_sync/ntp/ntp_basics#holdover_behavior_and_root_dispersion

> If the client has no other time source configured then it can keep accepting the drifting server anyway, so all clients of that server will at least keep the same time. If clients would immediately discard this server even if they had no other time source available then this would be even worse since even the times on different clients would start to drift apart.

I think this describes the heartbeat issue we observed. Our concern should be that the fabric is in sync (at least) internally and let the customer take care of the external synchronization.

PR #441, opened 2025-03-05
ci: use slugs and upload reg logs

Frostman:
It should make all uploaded artifacts sortable to group things by job/stage and easier to navigate.


pau-hedgehog (approved):

pau-hedgehog:
This looks way more organized!

Why do we store registry logs, btw?

Frostman:
@pau-hedgehog to be able to see if there was some communication to upstream issue or smth caused by the registry itself

PR #450, opened 2025-03-10
feat: frr external for vlab testing

edipascale:
if starting the vlab with the --externals option, a new "external" vm will be created representing the external peering device. It
runs FRR and it is autoconfigured from the wiring information.

Note that there are several caveats:
- there is no monitoring of changes to the wiring once the initial configuration step is completed
- on any given connection, you can have several vlan tagged attachments OR an untagged one, but not a mix of both, due to an issue where outgoing untagged taffic from the vm gets tagged with vlan 4095 if there are vlan subinterfaces on it
- since virtual switches do not support ACLs, peering multiple VPCs to the same external on a virtual switch will result in those VPCs being able to talk to each other, breaking connectivity tests; this is not a limitation of this patch

Closes https://github.com/githedgehog/internal/issues/78


pkg/hhfab/vlab_frr_butane.tmpl.yaml
```diff
@@ -0,0 +1,248 @@
...
+#          [Unit]
+#          Description=Reload FRRouting configuration
+#          Requisite=frr.service
+#          After=frr.service
+#          Type=oneshot
+#
+#          [Service]
+#          ExecStart=/usr/bin/docker exec frr /usr/lib/frr/frr-reload
+
+    - path: /usr/bin/vtysh
```
mrbojangles3, line 214: `/usr/bin` is a special place in flatcar. It doesn't like the binaries added there since that is a read-only file system. The binaries that we supply to flatcar are in /opt/hedgehog.

pkg/hhfab/vlab_frr_butane.tmpl.yaml
```diff
@@ -0,0 +1,248 @@
...
+           match community {{$vrfkey}}Out
+          !
+          route-map {{$vrfkey}}Out deny 2000
+          !{{end}}
+          line vty
+
+    - path: /etc/frr/vtysh.conf
+      mode: 0644
+      user:
+        id: 100
```
mrbojangles3, line 145: the default user "core" is uid/gid 500. I would change that in the file just in case that is a problem.

pkg/hhfab/vlab_frr_butane.tmpl.yaml
```diff
@@ -0,0 +1,248 @@
...
+      enabled: false
+      contents: |
+        [Unit]
+        Description=FRRouting Container
+        After=docker.service network-online.target
+        Requires=docker.service
+
+        [Service]
+        Type=simple
+        ExecStart=/usr/bin/docker run --name=frr --rm \
```
mrbojangles3, line 233: I also got rid of the multi-line ExecStart argument here. Again, not something ignition said I HAD to do but I was throwing some fixes at the wall.

pkg/hhfab/vlab_frr_butane.tmpl.yaml
edipascale: 100/100 is the uid/gid of the default "frr" user in the docker image; this was needed so that it can read the config files, else I get permission errors. I could try to change `core`'s uid/gid to match the expected one in the docker image, but maybe it's better/simpler to keep them separate

pkg/hhfab/vlabconfig.go
```diff
@@ -552,6 +621,112 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
 			}
+		} else if conn.Spec.External != nil && !cfg.ExtInvalid {
+			if externalID > MaxExternalConns {
+				slog.Warn("too many external connections", "max-external-connections", MaxExternalConns)
+				cfg.ExtInvalid = true
+
+				continue
+			}
+			// FIXME: this is a hack to get the external name and switch name from the connection
+			connSplits := strings.SplitN(conn.Name, "--", 3)
```
Frostman, line 632: The connection could be (and would be in the case of multiple externals) used by multiple externals, so it's better not to rely on it.

pkg/hhfab/vlabconfig.go
```diff
@@ -552,6 +621,112 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
+			if len(connSplits) != 3 {
+				slog.Warn("unexpected external connection name format, cannot derive external/switch", "connection", conn.Name)
+				cfg.ExtInvalid = true
+
+				continue
+			}
+			switchName := connSplits[0]
```
Frostman, line 639: The switch name could be extracted from the `spec.eternal.link.switch.port`

Frostman (changes requested):

pkg/hhfab/vlabconfig.go
edipascale: ack, I wrongly assumed it would be one connection per external. I now iterate over the external attachments using that connection instead

edipascale:
after troubleshooting this with @Frostman last week we realized there is something weird when using google's dns as the curl target on top of the external natting, and while we did not find the root cause, for now I will change the target to cloudflare's dns, which appears to work reliably.

Note that the PR is still marked as draft because I realised there is a problem with multiple external peerings: test-connectivity will fail because pings between the VPCs peered with the externals will go through it and succeed, regardless of vpc peerings. I must tune the iptables rules in the external so that traffic coming from the fabric cannot be forwarded back to it directly - i.e. without going through the internet-facing interface first.

pkg/hhfab/vlabrunner.go
```diff
@@ -325,6 +344,9 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 	postProcessDone := make(chan struct{})
 
 	for _, vm := range vlab.VMs {
+		if vm.Type == VMTypeExternal && !opts.SpawnExternal {
```
Frostman, line 347: We shouldn't be creating VLAB VM object if it's not needed

Frostman (changes requested):

pkg/hhfab/vlabrunner.go
```diff
@@ -39,9 +39,12 @@ import (
 	coreapi "k8s.io/api/core/v1"
 )
 
-//go:embed vlab_butane.tmpl.yaml
+//go:embed vlab_server_butane.tmpl.yaml
 var serverButaneTmpl string
 
+//go:embed vlab_external_butane.tmpl.yaml
+var externalButaneTmpl string
```
Frostman, line 46: suggested improvements https://github.com/githedgehog/fabricator/pull/450/commits/8ae07dc212dcfb92caff9de634e6be99f53fdaa2

pkg/hhfab/vlabconfig.go
```diff
@@ -574,9 +661,110 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
+				sw := &wiringapi.Switch{}
+				if err := wiring.Get(ctx, kclient.ObjectKey{Name: switchName, Namespace: conn.Namespace}, sw); err != nil {
+					return nil, fmt.Errorf("failed to get switch %q: %w", switchName, err)
+				}
+
+				// We get the external IP as the "neighbor IP" in the wiring, but we need to
+				// convert it to a prefix. Conversely, the fabric "switch IP" is given as a prefix
+				// and we need to convert it to an address for the BGP neighbor commands.
+				// So we take the prefix length of the switch IP and apply it to the external IP.
+				fabSwitchPrefix := netip.MustParsePrefix(extAttach.Spec.Switch.IP)
```
Frostman, line 716: Please, never use `Must` methods in a runtime functions, they'll cause panics if invalid input provided

pkg/hhfab/vlabconfig.go
```diff
@@ -574,9 +661,110 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
+				extAddr := netip.MustParseAddr(extAttach.Spec.Neighbor.IP)
```
Frostman, line 717: Same here

pkg/hhfab/vlabconfig.go
```diff
@@ -574,9 +661,110 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
+					extNicCfg.UntaggedCfg = attachCfg
+					untagged = true
+				} else {
+					extNicCfg.Attachments = append(extNicCfg.Attachments, attachCfg)
+					tagged = true
+				}
+				if tagged && untagged {
+					return nil, fmt.Errorf("both tagged and untagged attachments for the same external connection %s, this is not supported", conn.Name) //nolint:goerr113
+				}
+				slog.Info("Added attachment to external",
```
Frostman, line 740: let's make it a debug level please and let's make a message a bit more clear, like "Added attachment to virt external" or smth like this so it's clear that it's what happening to the virt external

pkg/hhfab/vlabconfig.go
```diff
@@ -414,6 +456,45 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
+			hwExternals[external.Name] = true
+
+			continue
+		}
+		if _, exists := cfg.Externals.VRFs[external.Name]; exists {
+			return nil, fmt.Errorf("duplicate external VRF name: %q", external.Name) //nolint:goerr113
+		}
+		asn := getAsn(&external)
+		if asn == 0 {
+			slog.Debug("external has no ASN annotation, will attempt to fetch it from the external attachments", "name", external.Name)
```
Frostman, line 486: Log message should start with capital letter and let's make message a bit more clear like "Virtual external ..."

pkg/hhfab/vlabconfig.go
```diff
@@ -574,9 +661,110 @@ func createVLABConfig(ctx context.Context, controls []fabapi.ControlNode, nodes
...
+				// nothing to do if the external is hardware
+				if hwExternals[extName] {
+					continue
+				}
+				extVrf, ok := cfg.Externals.VRFs[extName]
+				if !ok {
+					return nil, fmt.Errorf("external attachment %q has no associated VRF for the external %q", extAttach.Name, extName) //nolint:goerr113
+				}
+				if extVrf.ASN == 0 {
+					slog.Debug("Setting ASN for external", "external", extName, "attachment", extAttach.Name, "ASN", extAttach.Spec.Neighbor.ASN)
```
Frostman, line 700: Let's make msg little more clear by saying "... for virtual external"

pkg/hhfab/vlabconfig.go
```diff
@@ -72,17 +79,45 @@ type VM struct {
...
+}
+
 type VLABConfig struct {
-	SSHKey string              `json:"-"`
-	Sizes  VMSizes             `json:"sizes"`
-	VMs    map[string]VMConfig `json:"vms"`
+	SSHKey    string              `json:"-"`
+	Sizes     VMSizes             `json:"sizes"`
+	VMs       map[string]VMConfig `json:"vms"`
+	Externals ExternalsCfg        `json:"externals"`
```
Frostman, line 112: If you're saving ExternalsCfg, please make sure that all fields for all structs that are getting serialized has an explicit `json:"..."` tags (ExternalAttachCfg, ExternalNICCfg, ExternalVRFCfg, ExternalsCfg)

Frostman (changes requested):
A few minor things and we can get it merged

pkg/hhfab/testing.go
```diff
@@ -1163,7 +1163,7 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
 					}
 					client := clientR.(*goph.Client)
 
-					if err := checkCurl(ctx, opts, curls, serverA, client, "8.8.8.8", reachable); err != nil {
+					if err := checkCurl(ctx, opts, curls, serverA, client, "1.0.0.1", reachable); err != nil {
```
Frostman, line 1166: I'm very sorry for nitpicking, but could you please add a comment before this line about 8.8.8.8 not working for us for some reason and so switching to 1.0.0.1 as a workaroind

pkg/hhfab/vlabconfig.go
```diff
@@ -72,17 +79,45 @@ type VM struct {
 	Size       VMSize
 }
 
+type ExternalAttachCfg struct {
+	Prefix      string `json:"prefix"`       // IP prefix to configure on the NIC
+	NeighborIP  string `json:"neighbor_ip"`  // IP address of the BGP neighbor
```
Frostman, line 84: I'm very sorry for nitpicking, but we're using camelCase for json tags in other places (like K8s)

Frostman (approved):

PR #452, opened 2025-03-10
Release preps: bumps and misc

Frostman:


Frostman:
Just removing the ci.yaml changes to test the upgrades so merging with just builds being successful as 18/20 vlab-upgrades were successful, one failed to install old vlab and one failed due to panic in k8s cache.

PR #460, opened 2025-03-13
bump: bcm sonic 4.4.0 -> 4.4.1

Frostman:


Frostman:
It's technically incomplete because we also want to update the YANG models if there are any changes. As configuration is getting deployed, there seems to be no issues with backward compatibility with 4.4.0

Frostman:
It's complete with the agent version bump. It seems to be working fine with both 4.4.0 and 4.4.1

PR #426, opened 2025-03-04
Add retries to Server VM SSH

pau-hedgehog:
Solve SSH connection failures prevent connectivity tests to run and cause flakiness int the CI under some infra load issues

Issue #413, opened 2025-02-22
LLDP inspect expected neighbor port E1/32/3, got Ethernet126


pau-hedgehog:
https://github.com/githedgehog/fabricator/actions/runs/13454465323/job/37644846913

```
Switch: ds3000-01 (actual←→expected)

PORT   	CONNECTION                           	TYPE        	NEIGHBOR   	PORT                	DESCRIPTION                   	
E1/1/1 	ds3000-01--external--ds2000-02       	external    	ds2000-02←→	Ethernet2←→         	Hedgehog:                     	
       	                                     	            	           	                    	[control_vip=172.30.1.1/32]←→ 	
E1/1/2 	server-1--mclag--ds3000-01--ds3000-02	mclag       	server-1   	enp2s1              	                              	
E1/1/3 	server-2--mclag--ds3000-01--ds3000-02	mclag       	server-2   	enp2s1              	                              	
E1/1/4 	server-3--bundled--ds3000-01         	bundled     	server-3   	enp2s1              	                              	
E1/2/2 	server-3--bundled--ds3000-01         	bundled     	server-3   	enp2s2              	                              	
E1/17/1	ds3000-01--vpc-loopback              	vpc-loopback	ds3000-01  	E1/18/1             	Hedgehog Fabric               	
E1/17/2	ds3000-01--vpc-loopback              	vpc-loopback	ds3000-01  	E1/18/2             	Hedgehog Fabric               	
E1/18/1	ds3000-01--vpc-loopback              	vpc-loopback	ds3000-01  	E1/17/1             	Hedgehog Fabric               	
E1/18/2	ds3000-01--vpc-loopback              	vpc-loopback	ds3000-01  	E1/17/2             	Hedgehog Fabric               	
E1/28/1	ds4000-01--fabric--ds3000-01         	fabric      	ds4000-01  	E1/1/1              	Hedgehog Fabric               	
E1/29/1	ds4000-01--fabric--ds3000-01         	fabric      	ds4000-01  	E1/2/1              	Hedgehog Fabric               	
E1/30/1	ds4000-02--fabric--ds3000-01         	fabric      	ds4000-02  	E1/1/1              	Hedgehog Fabric               	
E1/31/1	ds4000-02--fabric--ds3000-01         	fabric      	ds4000-02  	E1/2/1              	Hedgehog Fabric               	
[...]

PR #462, opened 2025-03-14
Inspect retries only

pau-hedgehog:
This PR separates the inspect retry functionality from PR #426, which address #413, which we are hitting more frequently as it seems not related to the infra

### DIFF
--- a/pkg/hhfab/show-tech/control.sh
+++ b/pkg/hhfab/show-tech/control.sh
@@ -74,30 +74,22 @@ KUBECTL="/opt/bin/kubectl"
 # Githedgehog Resources
 # ---------------------------
 {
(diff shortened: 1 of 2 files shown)


pau-hedgehog:
Pending to address https://github.com/githedgehog/fabricator/pull/426#discussion_r1994413788

pau-hedgehog:
> Pending to address [#426 (comment)](https://github.com/githedgehog/fabricator/pull/426#discussion_r1994413788)

@Frostman I hope it makes sense now.

I am wondering if we should retry only the LLDP inspect and proceed with the rest only on success to fail faster and have to scroll up in case of failure. If LLDP inspect fails BGP inspect fails also with the current network architecture, right?

Frostman (approved):

PR #482, opened 2025-03-20
fix(testing): check if iperf server is up/reachable

edipascale:
and retry up to 10 times if not.

Fixes: #310

### DIFF
--- a/pkg/hhfab/testing.go
+++ b/pkg/hhfab/testing.go
@@ -1345,38 +1345,59 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
 	}
 
 	if err := iperfs.Acquire(ctx, 1); err != nil {
-		return fmt.Errorf("acquiring iperf semaphore: %w", err)
+		return fmt.Errorf("acquiring iperf3 semaphore: %w", err)
 	}
 	defer iperfs.Release(1)
 
 	ctx, cancel := context.WithTimeout(ctx, time.Duration(opts.IPerfsSeconds+30)*time.Second)
 	defer cancel()
 
-	slog.Debug("Running iperf", "from", from, "to", to)
+	slog.Debug("Running iperf3", "from", from, "to", to)
 
 	g, ctx := errgroup.WithContext(ctx)
 
 	g.Go(func() error {
 		out, err := toSSH.RunContext(ctx, fmt.Sprintf("toolbox -q timeout -v %d iperf3 -s -1", opts.IPerfsSeconds+25))
 		if err != nil {
-			return fmt.Errorf("running iperf server: %w: %s", err, string(out))
+			return fmt.Errorf("running iperf3 server: %w: %s", err, string(out))
 		}
 
 		return nil
 	})
 
 	g.Go(func() error {
-		time.Sleep(1 * time.Second) // TODO think about more reliable way to wait for server to start
-
+		// We could netcat to check if the server is up, but that will make the server shut down if
+		// it was started with -1, and if we don't add -1 it will run until the timeout, so change approach
+		time.Sleep(1 * time.Second)
+		var err error
+		var out []byte
 		cmd := fmt.Sprintf("toolbox -q timeout -v %d iperf3 -P 4 -J -c %s -t %d", opts.IPerfsSeconds+25, toIP.String(), opts.IPerfsSeconds)
-		out, err := fromSSH.RunContext(ctx, cmd)
-		if err != nil {
-			return fmt.Errorf("running iperf client: %w: %s", err, string(out))
+		maxRetries := 3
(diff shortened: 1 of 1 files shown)


pkg/hhfab/testing.go
```diff
@@ -1345,38 +1345,59 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
...
+		// We could netcat to check if the server is up, but that will make the server shut down if
+		// it was started with -1, and if we don't add -1 it will run until the timeout, so change approach
+		time.Sleep(1 * time.Second)
+		var err error
+		var out []byte
 		cmd := fmt.Sprintf("toolbox -q timeout -v %d iperf3 -P 4 -J -c %s -t %d", opts.IPerfsSeconds+25, toIP.String(), opts.IPerfsSeconds)
-		out, err := fromSSH.RunContext(ctx, cmd)
-		if err != nil {
-			return fmt.Errorf("running iperf client: %w: %s", err, string(out))
+		maxRetries := 10
```
Frostman, line 1375: 10 retries feel like really a lot... Aren't we going to hide some other issues this way?

edipascale: I'm not too concerned about hiding other issues as we will only retry on an ssh error (or, to be more precise, an error with the "ssh:" string in it), but I can lower it to something more reasonable. Does 3 sound better?

pkg/hhfab/testing.go
```diff
@@ -1345,38 +1345,59 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
...
 		}
 
 		return nil
 	})
 
 	g.Go(func() error {
-		time.Sleep(1 * time.Second) // TODO think about more reliable way to wait for server to start
-
+		// We could netcat to check if the server is up, but that will make the server shut down if
+		// it was started with -1, and if we don't add -1 it will run until the timeout, so change approach
```
Frostman, line 1370: Sorry, what this comment is about?

pkg/hhfab/testing.go
```diff
@@ -1345,38 +1345,59 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
...
-		if err != nil {
-			return fmt.Errorf("running iperf client: %w: %s", err, string(out))
+		maxRetries := 3
+		for retries := 0; retries < maxRetries; retries++ {
+			// Run iperf3 client
+			out, err = fromSSH.RunContext(ctx, cmd)
+			if err == nil {
+				break
+			}
+			// it doesn't look like the goph client defines error types to check with errors.Is
```
pau-hedgehog, line 1382: incomplete comment?

Frostman: looks complete, it's about `errors.Is`

pkg/hhfab/testing.go
pau-hedgehog: Correct me if I'm wrong but the retry logic wouldn't be masking Fabric issues as SSH goes through a separate network. So the retries are addressing the reliability in the test infrastructure itself

We should investigate anyhow why the server isn't ready when the client connects

pkg/hhfab/testing.go
edipascale: this PR originally took code from a closed PR by @pau-hedgehog that tried the netcat approach:
```
+                       g.Go(func() error {
+                               for retries := 0; retries < 10; retries++ {
+                                       cmd := fmt.Sprintf("nc -zv %s 5201", toIP.String())
+                                       out, err := fromSSH.RunContext(attemptCtx, cmd)
+                                       if err == nil {
+                                               slog.Debug("iperf3 server is ready", "server", to)
+                                               break
+                                       }
+                                       slog.Debug("iperf3 server not ready, retrying", "server", to, "retry", retries+1, "error", err, "output", string(out))
+                                       time.Sleep(1 * time.Second)
+                               }
+                               return nil
+                       })
```
But that was causing bad file descriptor errors on all clients, as the server shuts down after the nc probe, hence the comment. I thought I'd leave it for us to remember that the other approach did not work, but let me know if you think it is unnecessary or it needs reformulating

pkg/hhfab/testing.go
edipascale: If the retries work, I don't think there's anything to investigate - we are giving the server 1 second of head start, but under load in a virtualized environment there is no guarantee that it will be enough, that's why I thought retries would be sensible

pkg/hhfab/testing.go
pau-hedgehog: Oh, I think that was a very early attempt. I later tried with https://github.com/githedgehog/fabricator/pull/426 which got preempted with other tasks and I'm going to close in favor of this one

Frostman (approved):

PR #510, opened 2025-03-28
feat: check if bridge-nf-call-iptables!=0 which can break mgmt net

Frostman:


Frostman:
my worry is that if it's not causing issue all the time we could be over-protective, we should probably set some iptables rules instead

Issue #524, opened 2025-04-01
occasional test-connectivity failure with virtual external peerings


edipascale:
while testing the (still not merged) PR to autogenerate external wirings for the vlab I noticed flaky test connectivity under some circumstances.

Observed failures so far:
 - single external connection from leaf-01 (mclag)
 - single external attachment
 - curl failed when peering with vpc-01
 - curl succeeded when peering with vpc-02, vpc-07, vpc-03 (separately)

Things to understand:
- root cause
- mclag related or generic?

will update once I collect more data

edipascale:
this appears to be reliably reproducible. the problem is a routing divergence between the mclag switches.
When peering the external with vpc-01 (which has server-01, attached to leaf-01), everything looks good in leaf-01:
```
leaf-01# show ip route vrf VrfVvpc-01
Codes:  K - kernel route, C - connected, S - static, B - BGP, O - OSPF
        > - selected route, * - FIB route, q - queued route, r - rejected route
       Destination        Gateway                                                                    Dist/Metric   Last Update 
--------------------------------------------------------------------------------------------------------------------------------      
 S>*   0.0.0.0/0          via 172.30.96.201               Ethernet11.100                             1/0           00:01:41 ago
 C>*   10.0.1.0/24        Direct                          Vlan1001                                   0/0           00:23:06 ago
 C>*   172.30.96.200/31   Direct                          Ethernet11.100                             0/0           00:01:56 ago
leaf-01# ping vrf VrfVvpc-01 -I 10.0.1.1 1.0.0.1
PING 1.0.0.1 (1.0.0.1) from 10.0.1.1 VrfVvpc-01: 56(84) bytes of data.
64 bytes from 1.0.0.1: icmp_seq=1 ttl=253 time=19.9 ms
^C
--- 1.0.0.1 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 19.878/19.878/19.878/0.000 ms
```

However, leaf-02 does not get the static route:
```
leaf-02# show ip route vrf VrfVvpc-01
Codes:  K - kernel route, C - connected, S - static, B - BGP, O - OSPF
        > - selected route, * - FIB route, q - queued route, r - rejected route
       Destination        Gateway                                                                    Dist/Metric   Last Update 
--------------------------------------------------------------------------------------------------------------------------------      
[...]

Issue #522, opened 2025-04-01
Enable virtual external by default for VLAB and make on-ready setup/test-conn test it


Frostman:
We should probably extend setup-vpcs with creating a single VPC peering and single Ext peering so all the cases are covered and test-conn will not need adjustments

edipascale:
Due to the limitations of virtual switches it was decided to scrap this plan.

PR #529, opened 2025-04-03
vlab: generate wiring for externals

edipascale:
Generate wirings for externals as part of hhfab vlab gen. Add parameters to determine how many externals and external connections are created, and where the ext. connections are placed (i.e. mclag, eslag or orphan leaves).

All attachments are currently vlan tagged, I didn't want to over complicate things with extra params for untagged.
Also all externals are in the same namespace, and attachments are created for all externals and all connections.

Closes: #521

### DIFF
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -721,7 +721,7 @@ func Run(ctx context.Context) error {
 						Name:    "generate",
 						Aliases: []string{"gen"},
 						Usage:   "generate VLAB wiring diagram",
-						Flags:   flatten(defaultFlags, vlabWiringGenFlags),
+						Flags:   flatten(defaultFlags, vlabWiringGenFlags, []cli.Flag{yesFlag}),
 						Before:  before(false),
 						Action: func(_ *cli.Context) error {
 							builder := hhfab.VLABBuilder{
@@ -743,6 +743,7 @@ func Run(ctx context.Context) error {
 								ExtMCLAGConnCount:  uint8(wgExtMCLAGConns),  //nolint:gosec
 								ExtESLAGConnCount:  uint8(wgExtESLAGConns),  //nolint:gosec
 								ExtOrphanConnCount: uint8(wgExtOrphanConns), //nolint:gosec
+								YesFlag:            yes,
 							}
 
 							if err := hhfab.VLABGenerate(ctx, workDir, cacheDir, builder, hhfab.DefaultVLABGeneratedFile); err != nil {
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -173,6 +173,7 @@ func Run(ctx context.Context) error {
 	var wgMCLAGServers, wgESLAGServers, wgUnbundledServers, wgBundledServers uint
 	var wgNoSwitches bool
 	var wgGatewayUplinks uint
+	var wgExternals, wgExtMCLAGConns, wgExtESLAGConns, wgExtOrphanConns uint
 	vlabWiringGenFlags := []cli.Flag{
 		&cli.UintFlag{
 			Name:        "spines-count",
@@ -249,6 +250,26 @@ func Run(ctx context.Context) error {
 			Hidden:      !preview,
 			Destination: &wgGatewayUplinks,
 			Value:       2,
+		},
+		&cli.UintFlag{
+			Name:        "externals",
+			Usage:       "number of externals to generate",
+			Destination: &wgExternals,
+		},
+		&cli.UintFlag{
+			Name:        "external-mclag-connections",
+			Usage:       "number of external connections from MCLAG switches. NOTE: only 1 external connection in total is supported if using virtual switches",
+			Destination: &wgExtMCLAGConns,
+		},
+		&cli.UintFlag{
+			Name:        "external-eslag-connections",
+			Usage:       "number of external connections from ESLAG switches. NOTE: only 1 external connection in total is supported if using virtual switches",
+			Destination: &wgExtESLAGConns,
+		},
+		&cli.UintFlag{
+			Name:        "external-orphan-connections",
+			Usage:       "number of external connections from orphan switches. NOTE: only 1 external connection in total is supported if using virtual switches",
+			Destination: &wgExtOrphanConns,
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


edipascale:
I've added commits to also address #522, but we have a problem here:
- I've set the default to 1 external and 1 connection on an eslag leaf, similar to what we have in physical envs
- however, in collapsed-core mode only mclag switches are created, which means that the defaults are now invalid for this mode
- and on the other hand, we don't want to default to mclag externals because of #524 

In terms of having the ci passing, it would be easy enough to add explicit values to the vlab-gen for either the collapsed-core (if we keep the defaults as described above) or the spine-leaf cases (if we do not touch the current default). The question is what is the "least undesirable" option @Frostman

pkg/hhfab/vlabrunner.go
```diff
@@ -613,6 +615,24 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 
 						return fmt.Errorf("setting up VPCs: %w", err)
 					}
+				case OnReadySetupPeerings:
+					// TODO make it configurable
+					if err := c.SetupPeerings(ctx, vlab, SetupPeeringsOpts{
+						WaitSwitchesReady: true,
+						Requests:          []string{"1+4", "3~external-01:s=subnet-01"},
```
Frostman, line 622: Well, number of vpcs depends on the config as you can specify 0 servers and so no vpcs would be generated.

Frostman (changes requested):
We should probably generated conns to external from the the from both switches of the last redundancy group configured. Let's chat next week and discuss what would be the best approach.

pkg/hhfab/vlabbuilder.go
```diff
@@ -614,9 +667,71 @@ func (b *VLABBuilder) Build(ctx context.Context, l *apiutil.Loader, fabricMode m
 		}
 	}
 
+	if b.ExtCount > 0 {
+		externals := []vpcapi.External{}
+		extAsn := 64102
+		inboundAsnID := 65102
```
Frostman, line 673: ASN will depend on the specific leaf, but in case of communities those doesn't have to match the ASN, it's just a separate value, would be nice to rename to avoid misunderstanding

pkg/hhfab/vlabbuilder.go
```diff
@@ -614,9 +667,71 @@ func (b *VLABBuilder) Build(ctx context.Context, l *apiutil.Loader, fabricMode m
...
+		communityRuleID := 1000
+
+		for i := uint8(1); i <= b.ExtCount; i++ {
+			externalName := fmt.Sprintf("external-%02d", i)
```
Frostman, line 677: I know it sounds a bit ugly, but let's name the first one `default` so it would work with the `release-test` out of the box which is probably one of the most important goals

pkg/hhfab/vlabbuilder.go
```diff
@@ -728,3 +843,41 @@ func (b *VLABBuilder) createConnection(ctx context.Context, spec wiringapi.Conne
 
 	return conn, nil
 }
+
+func (b *VLABBuilder) createExternal(ctx context.Context, name string, spec vpcapi.ExternalSpec) (*vpcapi.External, error) {
+	external := &vpcapi.External{
+		TypeMeta: kmetav1.TypeMeta{
+			Kind:       "External", //?
```
Frostman, line 850: re comment - are you looking for a const for it?

cmd/hhfab/main.go
```diff
@@ -245,6 +246,28 @@ func Run(ctx context.Context) error {
...
+		&cli.UintFlag{
+			Name:        "external-mclag-connections",
+			Usage:       "number of external connections from MCLAG switches",
+			Destination: &wgExtMCLAGConns,
+		},
+		&cli.UintFlag{
+			Name:        "external-eslag-connections",
+			Usage:       "number of external connections from ESLAG switches",
+			Destination: &wgExtESLAGConns,
+			Value:       1,
```
Frostman, line 264: let's make it 2 by default so we're having redundant link

pkg/hhfab/vlabrunner.go
Frostman: I think the only safe thing would be to just start with the "1~default:s=subnet-01" (assuming first external is renamed to `default`. It's guaranteed that we'll have one VPC. We can improve it later.

Frostman (changes requested):

pkg/hhfab/vlabbuilder.go
edipascale: unfortunately we can't, because I'm using the external name as the name of the VRF in the virtual external VM and `default` is precisely the only reserved VRF name :) once we address the remaining issues I'll change the named external in the release test accordingly

pkg/hhfab/vlabbuilder.go
edipascale: exactly, I could not find it in fabric

pkg/hhfab/vlabrunner.go
edipascale: I'm fine with the change (minus the default naming for reasons explained below), but I assumed the on-ready commands were just a convenience hack to run hard-coded commands in the CI without spawning a new process, in which case this is never going to be called by anyone else... so does it really matter? Wouldn't you have the same problem for the onReady setup-vpcs if you had no servers? We could always hide the command if needed or make it configurable (and pass the appropriate parameters in the CI), but personally I'd just leave it as is

edipascale:
current status:
* the changes in release-test to use the first external have to be reverted or at least adapted; we cannot use virtual externals for most scenarios, as multiple external peerings (together with the limitation on virtual switches and ACLs) mean that VPCs will be able to ping each other even when they shouldn't. We can either revert the change, or fetch the annotations for externals and discard non-hardware ones.
* on top of that, there is https://github.com/githedgehog/internal/issues/145 which makes test-connectivity jobs with the new external peering fail in the ci

edipascale:
> current status:
> 
> * the changes in release-test to use the first external have to be reverted or at least adapted; we cannot use virtual externals for most scenarios, as multiple external peerings (together with the limitation on virtual switches and ACLs) mean that VPCs will be able to ping each other even when they shouldn't. We can either revert the change, or fetch the annotations for externals and discard non-hardware ones.

I went for the second option but it's not perfect. Technically the issue is not with virtual externals, but with virtual switches attached to any type of externals. Because right now we only have hardware externals connected to hardware switches and virtual externals connected to virtual switches, discarding virtual externals is the easy solution... but we probably want to revisit it.

> * on top of that, there is [troubleshoot virtual external peering with active connections on multiple leaves internal#145](https://github.com/githedgehog/internal/issues/145) which makes test-connectivity jobs with the new external peering fail in the ci

I've already got an answer from BCM on the case I opened requesting more info, which I provided; let's see if we can get unblocked there. In the meanwhile I reduced the autogenerated external connections from 2 to 1, which should work even with this strange issue.

Also I'm seeing several failures in the CI, both here and in other PRs. but they do not appear related to the changes. I'll dig a bit more tomorrow.

edipascale:
@Frostman this is technically ready to review now. I've removed the default external generation as discussed, and added warnings in case an option with multiple external connections is selected. The only thing missing is a run on a physical environment to check whether we need any additional fix for https://github.com/githedgehog/lab/issues/247, unsure whether we want to wait for that before tagging this as ready

PR #530, opened 2025-04-04
Misc release test fixes

edipascale:
* the 'Multi-Subnets with filtering' is using local peering and so can fail if subinterfaces are not supported, tag it accordingly
* the total failure count at the end of the test was wrong if some test had been skipped at execution time, due to us not resetting the error (used for the skip message) and also counting it as a failure
* curls were not being executed unless the extended flag was selected. this was intentional to reduce testing time, but it was the wrong decision, as it's the only way we actually test external peerings right now. enable them by default
* we were not importing the default prefix from externals, so we were only testing the fact that we could not curl (as expected)
  * note that the contrary is true now, i.e. we always import the default prefix. It is probably worth covering negative tests too, as well as importing specific prefixes and testing against those, but that can be done in a separate PR

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -221,7 +221,7 @@ jobs:
     needs:
       - test
       - builds
-    timeout-minutes: 120
+    timeout-minutes: 150
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


Frostman (approved):

edipascale:
@Frostman I've added some more release-test fixes that I had in the external-autogen branch, since that one might take a while longer to be merged and I'd like @pau-hedgehog to have these

Frostman (approved):

PR #532, opened 2025-04-04
fix(generate): remove unneeded SwitchGroup empty

pau-hedgehog:


Frostman:
I've been using it for checking connectivity test as it should detect remote peering on the empty switch group

pau-hedgehog:
> I've been using it for checking connectivity test as it should detect remote peering on the empty switch group

Then it's not unneeded. Closing

PR #543, opened 2025-04-09
refactor: diagram to use kube client and support live mode

Frostman:


cmd/hhfab/main.go
```diff
@@ -598,36 +593,27 @@ func Run(ctx context.Context) error {
...
+						Usage: "load resources from actually running API instead of the config file (fab.yaml and include/*)",
 					},
 				}),
 				Before: before(false),
 				Action: func(c *cli.Context) error {
-					format := strings.ToLower(c.String("format"))
+					format := diagram.Format(strings.ToLower(c.String("format")))
 					styleType := diagram.StyleType(c.String("style"))
-					if err := hhfab.Diagram(workDir, format, styleType); err != nil {
+					if err := hhfab.Diagram(ctx, workDir, cacheDir, c.Bool("live"), format, styleType); err != nil {
```
pau-hedgehog, line 616: Maybe just add the short flag `-l`. For the rest, this is awesome!

pau-hedgehog (approved):

cmd/hhfab/main.go
Frostman: Yeah, makes sense, could you add it? thx

PR #547, opened 2025-04-13
fix(reinstall): add support for ds5000

### LINKED ISSUE #546: Switch reinstall gets into rescue mode on ds5000
Image

pau-hedgehog:
in the zds5000, the ONIE Grub menu started with Rescue option preselected

also it takes longer to start ONIE install from eth0 because it tries with usb0

Fixes #546


pau-hedgehog:
@edipascale I'm removing this because it was unused. But I don't recall why we didn't use it in the end
```
set ONIE_CHAIN_INSTALL "*ONIE: Install OS"
```

pkg/hhfab/vlabhelpers_reinstall.exp
```diff
@@ -188,77 +187,84 @@ if { ! $POWER_RESET} {
...
-				log_message "EXP-ERR" "Lost connection while sending return: $err"
-				exit $ERROR_CONSOLE
-			}
-			
-			expect {
-				-ex "GNU GRUB" {
+	# Wait for menu to be fully rendered
+	expect -timeout 10 -ex "Press enter to boot the selected OS" {
+		log_message "EXP-DBG" "GRUB menu fully rendered"
+		sleep 1
```
edipascale, line 194: nit: maybe we don't need this anymore, I think the problem before was that we could be sending the home key before the menu was fully rendered and that's why we slept for a second. It's fine if you prefer to leave it though (unless grub has a default selection timer of 1 second or less, which hopefully should never be the case)

edipascale (changes requested):
thanks @pau-hedgehog, waiting for the grub menu to be fully rendered is the right thing to do. However the new code has some `eof` guards missing, there should be one for each `timeout` guard to make sure we detect console errors while waiting for some line

edipascale:
> @edipascale I'm removing this because it was unused. But I don't recall why we didn't use it in the end
> 
> ```
> set ONIE_CHAIN_INSTALL "*ONIE: Install OS"
> ```

In the first version I wrote I was explicitly trying to match this, but then I was getting timing issues on some switch families and I changed it to just going to the first option. I think we decided that it was safe to assume that while a user could change the order of entries in the first grub menu, it was reasonable to assume that the ONIE menu would always have the Install Os option as the first one

pau-hedgehog:
```
ubuntu@env-4-1:~/hhfab$ ./hhfab-reinstall vlab switch reinstall -yw
14:11:21 INF Hedgehog Fabricator version=v0.37.0-1-gf0ac2533-dirty-lA1406-pc
14:11:21 INF Wiring hydrated successfully mode=if-not-present
14:11:21 INF VLAB config loaded file=vlab/config.yaml
14:11:21 INF Reinstalling switches mode=hard-reset switches=[]
14:11:22 INF Power managing switches action=cycle switches=[]
14:11:22 INF Calling PDU API switch=as7712-03 psu=psu1 pduIP=192.168.88.7 outletID=2 action=cycle
14:11:22 INF Calling PDU API switch=as7712-04 psu=psu1 pduIP=192.168.88.7 outletID=1 action=cycle
14:11:22 INF Calling PDU API switch=s5232-03 psu=psu1 pduIP=192.168.88.7 outletID=4 action=cycle
14:11:23 INF Calling PDU API switch=s5248-06 psu=psu1 pduIP=192.168.88.7 outletID=3 action=cycle
14:11:23 INF Calling PDU API switch=s5248-07 psu=psu1 pduIP=192.168.88.7 outletID=5 action=cycle
14:11:23 INF Calling PDU API switch=s5248-08 psu=psu1 pduIP=192.168.88.7 outletID=6 action=cycle
14:11:23 INF Calling PDU API switch=zds4101-01 psu=psu1 pduIP=192.168.88.7 outletID=8 action=cycle
14:11:23 INF Calling PDU API switch=zds5000-01 psu=psu1 pduIP=192.168.88.7 outletID=7 action=cycle
14:11:24 INF Switches power cycled
14:18:01 INF Switch reinstalled successfully name=zds4101-01
14:18:20 INF Switch reinstalled successfully name=s5248-07
14:18:21 INF Switch reinstalled successfully name=s5248-06
14:18:28 INF Switch reinstalled successfully name=s5232-03
14:18:30 INF Switch reinstalled successfully name=s5248-08
14:19:57 INF Switch reinstalled successfully name=as7712-03
14:20:08 INF Switch reinstalled successfully name=as7712-04
14:20:19 INF Switch reinstalled successfully name=zds5000-01
14:20:19 INF All switches reinstalled successfully took=8m57.320461268s
```

edipascale (approved):

pkg/hhfab/vlabhelpers_reinstall.exp
Frostman: I think it could be taken care of separately if needed - @pau-hedgehog

PR #548, opened 2025-04-14
chore: bump flatcar to 4152.2.2

mrbojangles3:
closes #544

### DIFF
--- a/pkg/fab/versions.go
+++ b/pkg/fab/versions.go
@@ -34,8 +34,8 @@ var Versions = fabapi.Versions{
 		Ctl:            FabricatorVersion,
 		NodeConfig:     FabricatorVersion,
 		Pause:          "3.6", // wait image from k3s // TODO embed wait into node-config image?
-		ControlUSBRoot: "v4152.2.0-hh1",
-		Flatcar:        "v4152.2.0",
+		ControlUSBRoot: "v4152.2.2-hh1",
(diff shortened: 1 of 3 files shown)


pkg/fab/versions.go
```diff
@@ -34,7 +34,7 @@ var Versions = fabapi.Versions{
 		Ctl:            FabricatorVersion,
 		NodeConfig:     FabricatorVersion,
 		Pause:          "3.6", // wait image from k3s // TODO embed wait into node-config image?
-		ControlUSBRoot: "v4152.2.0-hh1",
+		ControlUSBRoot: "v4152.2.2-hh1",
 		Flatcar:        "v4152.2.0",
```
Frostman, line 38: That should be bumped too as well as in the VLAB portion and README for that package

Frostman (changes requested):

pkg/fab/versions.go
mrbojangles3: :facepalm: I need to update my list of places

Frostman: I've added a unit test for it some time ago so it'll fail if versions aren't in sync, so no worries - https://github.com/githedgehog/fabricator/actions/runs/14453816501/job/40532412748

you can run locally by `just test`

Frostman (approved):

Issue #549, opened 2025-04-15
hhnet - bonding driver xmit_hash_policy


mrbojangles3:
According to [the bonding driver docs](https://www.kernel.org/doc/html/latest/networking/bonding.html), 802.3ad has two different compliant hashing policies. `layer2` and `layer2+3`. Currently the default bond created by hhnet is a layer2 hashing type. We should test and ,if successufl, advertise the second hashing policy.

edipascale:
@mrbojangles3 is the idea to replace it always or to "expose it" somehow? I could easily add the option to hhnet, but right now in fabricator, to the best of my knowledge, this is only called implicitly as part of some vlab steps (specifically `up`, `setup-vpcs` and some of the relase-tests). We could expose it as an option for each/some of these but I'm not too sure that it makes sense, and I wanted to double check what was the ask here.

mrbojangles3:
My thought was that we would expose it and test it in a few places. My
reasoning is just that we use the spec compliant 802.3ad lacp setup, and
that offers to types of hashing. I figure our customers might want to use
either.

Logan Blyth

PR #553, opened 2025-04-16
chore: use layer2+3 hashing in hhnet bond

edipascale:
Part of #549


cmd/hhfab/main.go
```diff
@@ -881,6 +881,12 @@ func Run(ctx context.Context) error {
 								Aliases: []string{"mtu"},
 								Usage:   "interface MTU for VPCs advertised by DHCP",
 							},
+							&cli.StringFlag{
+								Name:    "hash-policy",
+								Aliases: []string{"hash"},
+								Usage:   "xmit_hash_policy for bond interfaces on servers",
```
Frostman, line 887: could you please list supported ones

pkg/hhfab/testing.go
```diff
@@ -327,7 +330,7 @@ func (c *Config) SetupVPCs(ctx context.Context, vlab *VLAB, opts SetupVPCsOpts)
 		if conn.Spec.Unbundled != nil {
 			netconfCmd = fmt.Sprintf("vlan %d %s", vlan, conn.Spec.Unbundled.Link.Server.LocalPortName())
 		} else {
-			netconfCmd = fmt.Sprintf("bond %d", vlan)
+			netconfCmd = fmt.Sprintf("bond %d %s", vlan, opts.HashPolicy)
```
Frostman, line 333: please make sure that valid value is being passed

cmd/hhfab/main.go
```diff
@@ -881,6 +881,12 @@ func Run(ctx context.Context) error {
 								Aliases: []string{"mtu"},
 								Usage:   "interface MTU for VPCs advertised by DHCP",
 							},
+							&cli.StringFlag{
+								Name:    "hash-policy",
+								Aliases: []string{"hash"},
+								Usage:   "xmit_hash_policy for bond interfaces on servers [layer2|layer2+3|layer3+4|encap2+3|encap3+4|vlan+srcmac]",
```
mrbojangles3, line 887:
```suggestion
								Usage:   "xmit_hash_policy for bond interfaces on servers [layer2|layer2+3]",
``` 
we only support the hashing policy that are 802.3ad compliant. Truth be told, I am not sure how / when the other policies are deployable, presumably when the switch and NIC are from the same manufacturer or testing has been done.

edipascale: as discussed offline, I've kept the other modes but added a warning if a non fully compliant option is selected. I've also exposed the hash-policy flag in the release-test so we can run tests with each of these and determine which ones are working (as part of https://github.com/githedgehog/fabricator/issues/559)

mrbojangles3 (approved):

mrbojangles3:
This change will also need to be in the docs example for vlab [here](https://docs.hedgehog.cloud/latest/vlab/demo/#setting-up-networking-on-test-servers) Mention this PR in a PR there?

Frostman (approved):

PR #603, opened 2025-05-09
feat(upgrade): specify upgrade starting version constraint

Frostman:


Frostman:
It's not working correctly due to the issue with the fabricator status versions calculation which shouldn't be touched in some cases like reading one before the upgrade but then it should get an override when we're applying the upgrade

PR #651, opened 2025-06-02
ci: security scans for control and gateway VMs

pau-hedgehog:
Adds GHA workflow to run trivy on control and gateway VMs and uploads SARIF scan results to Github

Findings can be seen here: https://github.com/githedgehog/fabricator/security/code-scanning?query=pr%3A651+is%3Aopen


Frostman (changes requested):
Do you intend to only run it in CI? (Not to do now) It's generally would be nice to make it available as a Just target but in that case it should be non-destructive - e.g. don't modify networking

pau-hedgehog:
> Do you intend to only run it in CI? (Not to do now) It's generally would be nice to make it available as a Just target but in that case it should be non-destructive - e.g. don't modify networking

The scripts can be run locally also. So a just target is feasible.

The network modification is a hack to grant gw vm internet access to run trivy. Is there a proper way?

pau-hedgehog:
> > Do you intend to only run it in CI? (Not to do now) It's generally would be nice to make it available as a Just target but in that case it should be non-destructive - e.g. don't modify networking
> 
> The scripts can be run locally also. So a just target is feasible.
> 
> The network modification is a hack to grant gw vm internet access to run trivy. Is there a proper way?

After discussing, this PR should be complete if I can run trivy in the Gateway in airgapped mode, without fiddling with the networking. I will investigate as it should be possible according to:

pau-hedgehog:
Closing in favor of https://github.com/githedgehog/fabricator/pull/670

PR #677, opened 2025-06-18
Fix collecting support dump when no GW

Frostman:


pkg/hhfab/vlabrunner.go
```diff
@@ -1105,6 +1105,13 @@ func (c *Config) vmPostProcess(ctx context.Context, vlab *VLAB, d *artificer.Dow
 
 				slog.Debug("K8s API on control node is ready", "vm", vm.Name, "type", vm.Type)
 			}
+
+			cmd := "PATH=/opt/bin kubectl hhfab support dump -yv"
```
edipascale, line 1109: nit: if the purpose of this is running the support dump in the CI for testing, wouldn't it be better to do so explicitly there instead of running it every time any vlab comes up?

edipascale (dismissed):

pkg/hhfab/vlabrunner.go
Frostman: It shouldn't add more then a few seconds but if it fails we should never ignore it so it felt like a good place with a side effect that we'll have a support dump in a home dir right after control node is installed just in case

PR #694, opened 2025-06-24
Add L3VNI VPC mode support

Frostman:


Frostman:
@edipascale I've returned back the remote peering for now, let's double check it

edipascale (approved):

PR #733, opened 2025-07-05
ci: refactor main workflow to make it reusable

pau-hedgehog:


.github/workflows/ci.yaml
```diff
@@ -214,407 +214,287 @@ jobs:
             exit 1
           fi
 
-  vlab:
+  # VLAB tests - separate job for each matrix combination
+  vlab-spine-leaf-usb:
```
Frostman: can we keep matrix jobs? they are very useful when changing permutations so no other changes needs to be done

.github/workflows/ci.yaml
```diff
@@ -214,407 +214,287 @@ jobs:
...
     runs-on: ubuntu-latest
     needs:
       - hlab
     if: ${{ always() }}
-
     steps:
       - run: |
-          result="${{ needs.hlab.result }}"
-          if [[ $result == "success" || $result == "skipped" ]]; then
+          if [[ "${{ needs.hlab.result }}" == "success" || "${{ needs.hlab.result }}" == "skipped" ]]; then
```
Frostman, line 364: the idea here was that hlab would become a matrix job at some point

.github/workflows/ci.yaml
```diff
@@ -214,407 +214,287 @@ jobs:
...
+      run_release_test: ${{ inputs.run_release_test == true || contains(github.event.pull_request.labels.*.name, 'release-test') }}
+      debug_enabled: ${{ inputs.debug_enabled }}
 
+  # Simplified check for matrix job results
   vlabs:
     runs-on: ubuntu-latest
     needs:
-      - vlab
-      - vlab-upgrade
+      - vlab-spine-leaf-usb
```
Frostman: `vlabs` could be used here

.github/workflows/ci.yaml
pau-hedgehog: This PR does not touch ci.yaml anymore

PR #762, opened 2025-07-14
Switch to reusable VLAB workflow 

### LINKED ISSUE #765: Reusable VLAB workflow

Frostman:
Fixes #765


edipascale (dismissed):
LGTM. a couple of nits which might be irrelevant or addressed separately:
* did we test that release-test label is still picked up correctly by the workflow?
* I would like the `l2vni` mode to be accepted in yamls too. I understand that we needed an empty default for backward compatibility but ideally we'd transition to an explicit default `l2vni` vpc mode, or at least accept both that and an empty string everywhere

PR #773, opened 2025-07-17
fix: container extraction from trivy scans

pau-hedgehog:
Fixed findings can be seen here: https://github.com/githedgehog/fabricator/security/code-scanning?query=pr%3A773+is%3Aopen

PR #792, opened 2025-07-21
Pau/ci trivy options

pau-hedgehog:


pau-hedgehog:
> It seems like it's still producing strange results, e.g. [golang.org/x/crypto/ssh: Misuse of ServerConfig.PublicKeyCallback may cause authorization bypass ...](/githedgehog/fabricator/security/code-scanning/1163)
> 
> it's pointing to `control-1/fabric:v0.84.3/zot-linux-amd64:1` which isn't possible as there is no zot in the fabric container.

Thanks for pointing this out. I'll address this in https://github.com/githedgehog/fabricator/pull/773, which is the one supposed to be fixing thit

Frostman (approved):

PR #794, opened 2025-07-22
Enable GW testing in CI

Frostman:


pkg/hhfab/release.go
```diff
@@ -867,8 +872,8 @@ func (testCtx *VPCPeeringTestCtx) multiSubnetsIsolationTest(ctx context.Context)
 			return errors.Join(err1, err2)
 		}
 		time.Sleep(5 * time.Second)
-		if err := WaitSwitchesReady(ctx, testCtx.kube, 1*time.Minute, 5*time.Minute); err != nil {
-			return fmt.Errorf("waiting for switches to be ready: %w", err)
+		if err := WaitReady(ctx, testCtx.kube, WaitReadyOpts{AppliedFor: waitAppliedFor, Timeout: waitTimeout}); err != nil {
```
edipascale, line 875: nit: for other Opts structs in the release tests, we create them at the start, save them to testCtx and pass them to the function calls when needed. I can make this change later if needed

pkg/hhfab/testing.go
```diff
@@ -84,19 +83,188 @@ func (c *Config) Wait(ctx context.Context, vlab *VLAB) error {
...
+		ctx, cancel = context.WithTimeout(ctx, opts.Timeout)
+		defer cancel()
 	}
-	slog.Info("All gateways are ready", "took", time.Since(start))
 
-	return nil
+	if opts.PollInterval == 0 {
+		opts.PollInterval = 5 * time.Second
+	}
+	if opts.PrintEvery == 0 {
```
edipascale, line 121: very minor nit: do we want to allow no printing at all, in which case 0 could be a valid choice that we shouldn't overwrite? Personally I would do that in the release tests for example, it needlessly bloats the logs

edipascale (dismissed):
LGTM

pkg/hhfab/testing.go
```diff
@@ -84,19 +83,188 @@ func (c *Config) Wait(ctx context.Context, vlab *VLAB) error {
...
+		}
+		for _, sw := range switches.Items {
+			expectedSwitches = append(expectedSwitches, sw.Name)
+		}
+	}
+	if len(expectedSwitches) > 0 {
+		slog.Info("Expected switches", "agent", expectedSwAgentV, "switches", expectedSwitches)
+	}
+
+	// gateway controller will set agent version to it's own version by default
```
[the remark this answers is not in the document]
Frostman: it's AI, it's not me!

pkg/hhfab/testing.go
Frostman: I've actually implemented it and reverted b/c the thing is that switch can loose "ready" state for some reason (agent restart, rogue update, etc) and we should actually wait till all of the switches and gateways are ready at the same time IMO

pkg/hhfab/release.go
edipascale: Agreed, I have an item on my mental list to try and tune down at least the appliedFor times in release tests, maybe removing them all together if it does not affect the results. They were set at 1 minute when we were struggling with a lot of instability in the CI, but things have generally improved now, and I don't know that this is even helping at all anyway

pkg/hhfab/testing.go
```diff
@@ -1983,6 +1983,12 @@ func checkIPerf(ctx context.Context, opts TestConnectivityOpts, iperfs *semaphor
 		return nil
 	}
 
+	iPerfsMinSpeed := opts.IPerfsMinSpeed
+	// TODO remove workaround after we have better GW performance
+	if reachability.Reason == ReachabilityReasonGatewayPeering {
```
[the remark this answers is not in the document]
edipascale: ha, that's a surprise! would checking that it's greater than 0 be sufficient to detect that issue? I just want to avoid going back to the situation that made us remove the check for virtual switches in the first place

PR #844, opened 2025-08-04
chore: diagram add external support

pau-hedgehog:


pkg/hhfab/diagram/mermaid.go
```diff
@@ -82,98 +80,100 @@ func generateMermaid(topo Topology) string {
...
-					}
-
-					mclagGroups[groupID] = append(mclagGroups[groupID], node)
-
-					for _, pairNode := range layers.Leaf {
-						if pairNode.ID == pair {
-							mclagGroups[groupID] = append(mclagGroups[groupID], pairNode)
-
-							break
+			if groupName, hasGroup := node.Properties["redundancyGroup"]; hasGroup && groupName != "" {
```
Frostman, line 127: @pau-hedgehog, it would be great to make consts for all that hardcoded magic strings to avoid issues like typos in the future

hack/diagrams.just
```diff
@@ -0,0 +1,181 @@
+# Generate diagrams for multiple environments in different formats and styles
+test-diagram vlab_workdir="" y="":
```
Frostman, line 2: it would be nice for it to depend on building the hhfab automatically bin as it relies on it

Frostman (approved):
@pau-hedgehog few comments, could you please fix it separately, I'll merge this now

PR #857, opened 2025-08-11
chore: add multi VPC attachment test

pau-hedgehog:


Frostman:
I think we shouldn't be doing it now, there are too many things to do:

- SetupVPCs is already overloaded, and to make it properly configurable, it would be difficult w/o rewriting and introducing some kind of simple lang to define VPCs (similar to the peerings)
- We cannot rely on `bond0` - what if you have multiple multihomed connections at the same time
- We should calculate source interface for pings/iperfs/curls very carefully and it technically could be a list to test, not just a single one
- When we're doing multi-attach to servers we would need to test all permutations - check connectivity from all attachments of all servers to all attachments of all servers to make sure that traffic will not go through where we don't expect it

PR #869, opened 2025-08-12
feat: change to tinyproxy

mrbojangles3:


api/fabricator/v1beta1/fabricator_types.go
```diff
@@ -93,7 +93,7 @@ type ComponentsStatus struct {
 	FabricCtrl           ComponentStatus `json:"fabricCtrl,omitempty"`
 	FabricBoot           ComponentStatus `json:"fabricBoot,omitempty"`
 	FabricDHCP           ComponentStatus `json:"fabricDHCP,omitempty"`
-	FabricProxy          ComponentStatus `json:"fabricProxy,omitempty"`
+	ControlProxy         ComponentStatus `json:"controlProxy,omitempty"`
```
Frostman, line 96: Let's move it to the

pkg/fab/comp/controlproxy/controlproxy.go
```diff
@@ -0,0 +1,90 @@
...
+		urls = append(urls, u.Hostname())
+	}
+	for _, val := range cfg.Spec.Config.Fabric.DefaultAlloyConfig.LokiTargets {
+		u, err := url.Parse(val.URL)
+		if err != nil {
+			return nil, fmt.Errorf("url parsing loki target failed: %w", err)
+		}
+		urls = append(urls, u.Hostname())
+		urls = slices.Compact(urls)
+	}
```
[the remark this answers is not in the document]
Frostman: Yeah, Copilot is correct here. You should use `lo.Uniq` instead from `"github.com/samber/lo"`.

config/helm/control-proxy/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+fullNameOverride: "control-proxy"
+
+# This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
+image:
+  repository: ghcr.io/githedgehog/fabricator/control-proxy
+  # This sets the pull policy for images.
+  pullPolicy: IfNotPresent
+
+#This section builds out the service account more information can be found here: https://kubernetes.io/docs/concepts/security/service-accounts/
+serviceAccount:
```
Frostman, line 17: We don't need service account at all, let's drop it

config/helm/control-proxy/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
+TinyproxyConfig:
```
Frostman, line 46: It looks like it should be just a multiline string, e.g.:

```
config: |
   asd
   asd
```

Frostman: Additionally, it's a tineproxy chart so no need to prefix key with word "tinyproxy" and keys should be named in lowerCamelCase

config/helm/control-proxy/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+  - User 65532
+  - Group 65532
+  - Port 3128
+  - LogLevel Connect
+  - MaxClients 1024
+  - Filter "/conf/filters.conf"
+  - FilterDefaultDeny Yes
+  - FilterType fnmatch
+
+tinyproxyURLs: []
```
Frostman, line 56: ditto

Frostman (changes requested):
A few minor comments, everything else looks good. It'll currently break metrics collection for gateway if it'll be using endpoints not used for switches but it's okay, that chunk of code needs a redesign anyways.

pkg/fab/comp/controlproxy/controlproxy.go
```diff
@@ -0,0 +1,90 @@
...
+			return nil, fmt.Errorf("url parsing prometheus target failed: %w", err)
+		}
+		urls = append(urls, u.Hostname())
+	}
+	for _, val := range cfg.Spec.Config.Fabric.DefaultAlloyConfig.LokiTargets {
+		u, err := url.Parse(val.URL)
+		if err != nil {
+			return nil, fmt.Errorf("url parsing loki target failed: %w", err)
+		}
+		urls = append(urls, u.Hostname())
```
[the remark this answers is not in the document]
Frostman: Not sure if it's true, but worse checking

PR #870, opened 2025-08-12
Enable gateway peering testing

Frostman:


edipascale (approved):
LGTM, although Copilot's nit seems reasonable enough to address it

Frostman:
@edipascale those are two different values - min VS speed and min GW speed

PR #916, opened 2025-08-27
Cleanup peerings on setup-vpc and short on-ready cmds 

Frostman:


pkg/hhfab/testing.go
```diff
@@ -458,6 +459,18 @@ func (c *Config) SetupVPCs(ctx context.Context, vlab *VLAB, opts SetupVPCsOpts)
...
+
+		if err := client.IgnoreNotFound(kube.DeleteAllOf(ctx, &vpcapi.VPCPeering{})); err != nil {
+			return fmt.Errorf("cleaning up vpc peerings: %w", err)
+		}
+		if c.Fab.Spec.Config.Gateway.Enable {
+			if err := client.IgnoreNotFound(kube.DeleteAllOf(ctx, &gwapi.Peering{})); err != nil {
+				return fmt.Errorf("cleaning up gateway peerings: %w", err)
+			}
+		}
+	}
```
edipascale, line 473: can we add external peerings too? then we won't need to call wipe at all

Frostman: that makes sense

edipascale (approved):

PR #923, opened 2025-09-01
chore: add DHCP renew release test

pau-hedgehog:


pkg/hhfab/release.go
```diff
@@ -2273,6 +2274,210 @@ func (testCtx *VPCPeeringTestCtx) breakoutTest(ctx context.Context) (bool, []Rev
...
+}
+
+func (testCtx *VPCPeeringTestCtx) performDHCPRenewal(serverName, ifName string) error {
+	if err := execNodeCmd(testCtx.hhfabBin, testCtx.workDir, serverName,
+		fmt.Sprintf("sudo networkctl renew %s", ifName)); err != nil {
+		return fmt.Errorf("networkctl renew failed: %w", err)
+	}
+
+	// Wait for renewal to complete with timeout
+	ctx, cancel := context.WithTimeout(context.Background(), DHCPRenewalMaxWait)
```
edipascale, line 2460: you should be passing the existing `ctx` to this function and using it to create the timeout context, rather than using a new background context, so that if the overall context is terminated this one will too

pkg/hhfab/release.go
```diff
@@ -2273,6 +2274,210 @@ func (testCtx *VPCPeeringTestCtx) breakoutTest(ctx context.Context) (bool, []Rev
...
+	defer cancel()
+
+	// Simple validation that the interface still exists and is configured
+	for {
+		out, err := execNodeCmdWOutput(testCtx.hhfabBin, testCtx.workDir, serverName,
```
edipascale, line 2465: should we check the lease time to be more precise and make sure that the DHCP renewal actually succeeded? there is already some code doing it in the regular DHCP test that you could reuse

pkg/hhfab/release.go
```diff
@@ -2273,6 +2274,210 @@ func (testCtx *VPCPeeringTestCtx) breakoutTest(ctx context.Context) (bool, []Rev
...
+	Interface string
+}
+
+type RenewalResult struct {
+	Server   string
+	Duration time.Duration
+	Error    error
+}
+
+func (testCtx *VPCPeeringTestCtx) getServersWithVPCAttachments(ctx context.Context) ([]ServerWithInterface, error) {
```
edipascale, line 2396: nit: the name of the function is misleading, you're looking for the interface name rather than the attachment

pkg/hhfab/release.go
pau-hedgehog: As usual, you nailed. I realized that the networkctl option that indeed triggers renewal is not `renew`, not `forcerenew` but `reconfigure`

I'm adding the lease time checking and, as a bonus, for L3VNI checking the 2-step lease

I'm troubleshooting now some slow DHCP leases before resolving this. In case I'm too restrictive in the timeouts in the test or we have performance issues in the server

pkg/hhfab/release.go
edipascale: my preference would have been to keep it - I think it's better to have small reusable functions than a large chunk of code that becomes hard to follow - but it's not a big deal either way, we can leave it the way it is now. At some point we'll need to restructure this file and split it into multiple ones anyway, it's becoming huge

edipascale:
So I'm a little conflicted about this test the way it currently is, because I do not believe it's testing DHCP renewals anymore. Instead we are reconfiguring the interfaces and getting brand new leases. It's not a bad thing to have additional coverage on this, but it's not what we're saying we are doing, right? For instance, l3vni _actual_ renewals should not have the short->long lease cycle, or at least that's not what I would expect. I think to test renewals you need to set the lease to a reasonably short time via the DHCPOptions and wait long enough for it to expire, then check what you get afterwards. It might be possible to force a renew earlier via networkctl, although [this question](https://askubuntu.com/questions/1171393/renew-dhcp-lease-in-systemd-networkd) appears to imply hat it doesn't really work that way, and your experience seems to confirm it.

It's also easy to break this if we ever change the lease time values we use in the DHCP server. This is inevitable to a certain extent, although we could at least set the lease time as a DHCPOption for the "large lease" value instead of using the default and risking it changing.

None of the above means that we should reject the PR, but I think it's worth thinking about what we're doing.

The one thing I think we should change regardless is the code to parse the DHCP lease - instead of duplicating it let's use the existing function or adapt it if it doesn't fit your needs.

pkg/hhfab/release.go
```diff
@@ -3105,6 +3106,460 @@ func (testCtx *VPCPeeringTestCtx) prometheusObservabilityTest(ctx context.Contex
...
+
+func (testCtx *VPCPeeringTestCtx) waitForDHCPRenewal(ctx context.Context, serverName, ifName string, shortLeaseTime uint32) error {
+	isL3Mode := testCtx.setupOpts.VPCMode == vpcapi.VPCModeL3VNI || testCtx.setupOpts.VPCMode == vpcapi.VPCModeL3Flat
+
+	ssh, err := testCtx.getSSH(ctx, serverName)
+	if err != nil {
+		return fmt.Errorf("getting ssh config for server %s: %w", serverName, err)
+	}
+
+	_, _, err = ssh.Run(ctx, fmt.Sprintf("sudo networkctl reconfigure %s", ifName))
```
edipascale: nit: I don't know if this produces any stderr output in case of failure, but if so we could add it to the error message to ease troubleshooting. when in doubt I generally add it

pau-hedgehog:
Tried to address all comments. Let's see now. Thanks as usual @edipascale

Frostman (approved):
@pau-hedgehog added a few comments, but doesn't block merging it, thank you

PR #946, opened 2025-09-08
Minor improvements to release test framework

pau-hedgehog:
Cosmetic improvements to make release tests more readable to human beings

### DIFF
--- a/pkg/hhfab/release.go
+++ b/pkg/hhfab/release.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -22,6 +23,7 @@ import (
 	vpcapi "go.githedgehog.com/fabric/api/vpc/v1beta1"
 	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
 	"go.githedgehog.com/fabric/pkg/util/pointer"
+	fabapi "go.githedgehog.com/fabricator/api/fabricator/v1beta1"
 	"go.githedgehog.com/fabricator/pkg/fab"
 	gwapi "go.githedgehog.com/gateway/api/gateway/v1alpha1"
 	"golang.org/x/sync/errgroup"
@@ -3076,7 +3078,8 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOpts 
 		}
 	}
 
-	// Create all suites for planning purposes
+	// Note: Creating test contexts early for suite planning. While this creates
+	// some unused instances, it keeps the code simple and maintains consistency
 	noVpcSuite := makeNoVpcsSuiteRun(&VPCPeeringTestCtx{})
 	singleVpcSuite := makeVpcPeeringsSingleVPCSuite(&VPCPeeringTestCtx{})
 	multiVpcSuite := makeVpcPeeringsMultiVPCSuiteRun(&VPCPeeringTestCtx{})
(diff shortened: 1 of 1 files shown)


pkg/hhfab/release.go
```diff
@@ -3045,45 +3076,86 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOpts
...
+	}
+
+	// Calculate cumulative offsets for progress tracking
+	cumulativeOffset := 0
 
 	noVpcTestCtx := makeTestCtx(kube, setupOpts, workDir, cacheDir, true, rtOpts, roceLeaves)
 	noVpcTestCtx.noSetup = true
-	noVpcSuite := makeNoVpcsSuiteRun(noVpcTestCtx)
-	noVpcResults, err := selectAndRunSuite(ctx, noVpcTestCtx, noVpcSuite, regexesCompiled, rtOpts.InvertRegex, skipFlags)
+	noVpcTestCtx.totalTests = totalRunnableTests
```
edipascale, line 3126: I should probably refactor this code as there's no real reason to create a new testCtx every time, we are only changing a handful of params... but until we do that, can we add these new variables as parameters to makeTestCtx, so we don't have to update them manually every time?

pkg/hhfab/release.go
```diff
@@ -3045,60 +3078,439 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOtps
...
+	mgmt := []string{}
+
+	for _, sw := range swList.Items {
+		switch {
+		case sw.Spec.Role.IsSpine():
+			spines = append(spines, sw.Name)
+		case sw.Spec.Role.IsLeaf():
+			leaves = append(leaves, sw.Name)
+		default:
+			// For any other roles, we can check the string representation
```
Frostman, line 3232: There are no other roles possible then the ones defined by SwitchRole - https://github.com/githedgehog/fabric/blob/83d69b2d4de3e2c3a01d2f201ac5e14ced2cecf6/api/wiring/v1beta1/switch_types.go#L38-L43 so please use it instead of contains check

pkg/hhfab/release.go
```diff
@@ -3045,60 +3078,439 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOtps
...
+			roleStr := string(sw.Spec.Role)
+			if strings.Contains(strings.ToLower(roleStr), "border") {
+				borders = append(borders, sw.Name)
```
Frostman, line 3235: it's most probably going to be mixed always, we don't have dedicated borders in any envs

pkg/hhfab/release.go
```diff
@@ -3045,60 +3078,439 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOtps
...
+			} else if strings.Contains(strings.ToLower(roleStr), "mgmt") {
```
Frostman, line 3236: We don't have a concept of mgmt switch

pkg/hhfab/release.go
```diff
@@ -3045,60 +3078,439 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOtps
...
+		}
+	}
+
+	// Switches by role - convert if-else to switch statement
+	spines := []string{}
+	leaves := []string{}
+	borders := []string{}
+	mgmt := []string{}
+
+	for _, sw := range swList.Items {
```
Frostman, line 3225: It'll be probably better just to iterate over the list of switches and print them one by one with the role and any other metadata needed and then the same things for the servers/etc, but to be honest why do we need that? We have it all in the bundle for the job with versions/wirings/etc

pkg/hhfab/release.go
```diff
@@ -3045,60 +3078,439 @@ func RunReleaseTestSuites(ctx context.Context, workDir, cacheDir string, rtOtps
...
+	// External connections
+	externals := &vpcapi.ExternalList{}
+	if err := kube.List(ctx, externals); err == nil && len(externals.Items) > 0 {
+		slog.Info("  External Connections:", "count", len(externals.Items))
+	}
+
+	slog.Info("───────────────────────────────────────────────────────────────────────────")
+
+	// Component Versions
+	slog.Info("Component Versions:")
```
Frostman, line 3301: That code will become outdated very quickly, we already have versions.txt in the CI bundle, and if you run it manually you can always run `hhfab versions` instead of having a code that'll go stale quickly.

Frostman (changes requested):

PR #947, opened 2025-09-09
refactor test-connectivity errors

edipascale:
define some structured errors so we can do more in-depth parsing if we ever want to; attempt to improve the readability of the error logs; and make sure to join all of the observed errors so we do not have to rely on logs to understand what went wrong.


edipascale:
Example output from release test with ping errors:
```
13:29:11 ERR FAIL test="Gateway Peering" error="testing gateway peering connectivity: ping server-09 -> server-08: unexpected ping result (sent 3, rcvd 2, expected true); ping server-08 -> server-09: unexpected ping result (sent 3, rcvd 2, expected true)"
```

Haven't caught an iperf or curl error yet

pkg/hhfab/testing.go
```diff
@@ -1975,48 +2031,84 @@ func checkPing(ctx context.Context, opts TestConnectivityOpts, pings *semaphore.
...
+	if pe.Sent == 0 {
+		pe.Msg = "cannot parse ping output to get sent packets"
+
+		return pe
+	}
 
-	pingOk := err == nil && strings.Contains(out, "0% packet loss")
-	pingFail := err != nil && strings.Contains(out, "100% packet loss")
+	pingOk := err == nil && pe.Sent == pe.Received
+	pingFail := err != nil && pe.Received == 0
```
Frostman, line 2061: should it be `pe.Received < pe.Sent` when only some pings lost?

pkg/hhfab/testing.go
```diff
@@ -1975,48 +2031,84 @@ func checkPing(ctx context.Context, opts TestConnectivityOpts, pings *semaphore.
...
 
 	// TODO better logging and handling of errors
 	slog.Debug("Ping result", "from", from, "to", to,
 		"expected", expected, "ok", pingOk, "fail", pingFail, "err", err, "out", out)
 
 	if pingOk == pingFail {
 		if err != nil {
-			return fmt.Errorf("running ping: %w: %s", err, out) // TODO replace with custom error?
+			pe.Msg = fmt.Sprintf("%s: %s", err, out) // TODO replace with custom error?
```
Frostman, line 2069: I guess it's time to remove the TODO msg :)

pkg/hhfab/testing.go
```diff
@@ -1582,75 +1624,82 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
...
+			numIperfErrs++
+		case *CurlError:
+			numCurlErrs++
+		}
+		err = errors.Join(err, e)
 	}
 
-	slog.Info("All connectivity tested successfully", "took", time.Since(start))
+	if err != nil {
+		slog.Error("errors in test-connectivity", "ping errors", numPingErrs, "iperf errors", numIperfErrs, "curl errors", numCurlErrs)
```
Frostman, line 1697: Log msg should start with capital letter and you can shorten the keys as you already saying `Errors in ...`, e.g. `ping errors` => `ping` and it'll be great to keep the `took` portion in case of fails too

I'd suggest to probably align fail/success pass to smth like

`Test connectivity failed` and `Test connectivity passed`

pkg/hhfab/testing.go
edipascale: that case is the "unexpected ping result" case of `pingOk == pingFail` - i.e. both are false. the existing code was already covering it and I did not touch it, but I can make the logic explicit if you prefer?

Frostman: Yeah, you're right, pingFail is used to check when ping is completely failed so we know that there is no connectivity

pkg/hhfab/testing.go
edipascale: done. I've also added the errors themselves to the error case - we want to log them in case you are running test-connectivity outside of the release tests

Frostman (approved):

Issue #928, opened 2025-09-03
More combinations of peerings release tests for GW


Frostman:
- loop of peerings
- mixed fabric/gateway peerings using same VPCs

PR #952, opened 2025-09-10
chore(testing): add more gateway peering tests

pau-hedgehog:
Adds coverage to gw peering testing:
- loop of gateway peerings
- mixed fabric/gateway peerings using same VPCs

Addresses: https://github.com/githedgehog/fabricator/issues/928

### DIFF
--- a/pkg/hhfab/release.go
+++ b/pkg/hhfab/release.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/hhfab/release.go
```diff
@@ -2333,6 +2333,147 @@ func (testCtx *VPCPeeringTestCtx) gatewayPeeringTest(ctx context.Context) (bool,
...
+	if err := testCtx.kube.List(ctx, vpcs); err != nil {
+		return false, nil, fmt.Errorf("listing VPCs: %w", err)
+	}
+	if len(vpcs.Items) < 4 {
+		return true, nil, fmt.Errorf("not enough VPCs for mixed peering chain test (need at least 4)") //nolint:goerr113
+	}
+
+	vpcPeerings := make(map[string]*vpcapi.VPCPeeringSpec, 0)
+	externalPeerings := make(map[string]*vpcapi.ExternalPeeringSpec, 0)
+	gwPeerings := make(map[string]*gwapi.PeeringSpec, 0)
```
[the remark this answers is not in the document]
pau-hedgehog: not consistent with existing pattern

pkg/hhfab/release.go
```diff
@@ -2333,6 +2323,148 @@ func (testCtx *VPCPeeringTestCtx) gatewayPeeringTest(ctx context.Context) (bool,
...
+	vpcPeerings := make(map[string]*vpcapi.VPCPeeringSpec, 0)
+	externalPeerings := make(map[string]*vpcapi.ExternalPeeringSpec, 0)
+	gwPeerings := make(map[string]*gwapi.PeeringSpec, len(vpcs.Items))
+
+	// Create loop: VPC0↔VPC1↔VPC2↔...↔VPCn-1↔VPC0
+	for i := 0; i < len(vpcs.Items); i++ {
+		vpc1 := vpcs.Items[i]
+		vpc2 := vpcs.Items[(i+1)%len(vpcs.Items)] // wrap around to create loop
+
+		vpc1Subnets := make([]string, 0, len(vpc1.Spec.Subnets))
```
edipascale, line 2351: it's unfortunate that we need to be explicit for gateway peerings and list all of the subnets of each VPC, rather than just passing an empty array like we do for normal vpc peering. Maybe it would make sense for the gateway peering helper function to take the actual VPC objects (rather than their names) and a list of optional subnet names, and if these are empty it would add all of the subnets of each VPC, so we do not have to copy+paste these loops everywhere?

pkg/hhfab/release.go
```diff
@@ -2333,6 +2323,148 @@ func (testCtx *VPCPeeringTestCtx) gatewayPeeringTest(ctx context.Context) (bool,
...
+		}
+
+		appendGwPeeringSpecByName(gwPeerings, vpc1.Name, vpc2.Name, vpc1Subnets, vpc2Subnets)
+	}
+
+	if err := DoSetupPeerings(ctx, testCtx.kube, vpcPeerings, externalPeerings, gwPeerings, true); err != nil {
+		return false, nil, fmt.Errorf("setting up gateway loop peerings: %w", err)
+	}
+
+	reverts := []RevertFunc{
```
edipascale, line 2368: same comment as for the other revert

PR #973, opened 2025-09-16
chore: add release-test list option

edipascale:
as requested by @mrbojangles3 (but I was also thinking about this for a while)


Frostman (approved):

Frostman:
@edipascale seems like it panics in CI

edipascale:
> @edipascale seems like it panics in CI

should be fixed now 🤞🏼

PR #980, opened 2025-09-18
Add release tests for NAT

pau-hedgehog:


pkg/hhfab/release.go
```diff
@@ -2603,9 +2605,746 @@ func appendGwExtPeeringSpec(gwPeerings map[string]*gwapi.PeeringSpec, vpc1 *vpca
...
+	// Initialize reverts slice early
+	reverts := []RevertFunc{}
+
+	// Dynamically discover any server's attachment to vpc-02 by listing all and filtering
+	allAttList := &vpcapi.VPCAttachmentList{}
+	if err := testCtx.kube.List(ctx, allAttList); err != nil {
+		return false, reverts, fmt.Errorf("listing VPCAttachments: %w", err)
+	}
+
+	var attachmentToRemove string
```
edipascale: definitely let's avoid parsing based on names here and use the existing API instead, e.g. list vpcattachments with labels for your target VPC etc.

pkg/hhfab/release.go
```diff
@@ -2603,9 +2605,746 @@ func appendGwExtPeeringSpec(gwPeerings map[string]*gwapi.PeeringSpec, vpc1 *vpca
...
+		ObjectMeta: kmetav1.ObjectMeta{
+			Name:      newVPCName,
+			Namespace: kmetav1.NamespaceDefault,
+		},
+		Spec: vpcapi.VPCSpec{
+			IPv4Namespace: testIPNSName, // Different namespace from vpc-01 - allows overlap
+			VLANNamespace: "default",
+			Subnets: map[string]*vpcapi.VPCSubnet{
+				"subnet-01": {
+					Subnet:  existingSubnetCIDR, // Same subnet as vpc-01! (10.0.1.0/24)
```
edipascale: same comment as above re the IP, comment is misleading

pkg/hhfab/release.go
```diff
@@ -2603,9 +2605,746 @@ func appendGwExtPeeringSpec(gwPeerings map[string]*gwapi.PeeringSpec, vpc1 *vpca
...
+					VLAN:    1020,               // Use a known available VLAN
```
edipascale: if in the future we have 20 VPCs on a setup with more servers we will clash; I'd pick something higher within the allowed space to stay on the safe side

pkg/hhfab/release.go
```diff
@@ -2603,9 +2605,746 @@ func appendGwExtPeeringSpec(gwPeerings map[string]*gwapi.PeeringSpec, vpc1 *vpca
...
+					Gateway: existingGateway,    // Same gateway IP as vpc-01
+					DHCP: vpcapi.VPCDHCP{
+						Enable: true,
+						Range: &vpcapi.VPCDHCPRange{
+							Start: dhcpStart, // 10.0.1.200 - overlapping with vpc-01's range
```
edipascale: same comment again

pkg/hhfab/rt_utils.go
```diff
@@ -588,12 +588,26 @@ func appendGwExtPeeringSpec(gwPeerings map[string]*gwapi.PeeringSpec, vpc1 *vpca
...
+	// VPC1NATCIDR specifies the NAT CIDR ranges for VPC1
+	VPC1NATCIDR []string
+	// VPC2NATCIDR specifies the NAT CIDR ranges for VPC2
+	VPC2NATCIDR []string
+	// StatefulNAT enables stateful NAT instead of stateless (default: false = stateless)
+	StatefulNAT bool
+}
+
+// appendGwPeeringSpec adds a single gateway peering spec to an existing map, which will be the input for DoSetupPeerings
+func appendGwPeeringSpec(gwPeerings map[string]*gwapi.PeeringSpec, vpc1, vpc2 *vpcapi.VPC, opts GwPeeringOptions) {
```
edipascale: if you make opts a pointer you can just pass nil when not needed instead of having to manually pass an empty struct as you are currently doing

pkg/hhfab/rt_nat_tests.go
```diff
@@ -0,0 +1,713 @@
...
+		return fmt.Errorf("NAT iperf3 test failed with %d errors: %v", len(errors), errors) //nolint:err113
+	}
+
+	slog.Info("NAT connectivity test (ping+iperf3) completed successfully", "took", time.Since(startTime))
+
+	return nil
+}
+
+// Test gateway peering with stateful source NAT (only VPC1 has stateful NAT configured)
+// NOTE: Stateful NAT on both sides of a peering is not supported (see dataplane#1248)
```
edipascale: I believe it is now (or will soon)

pkg/hhfab/rt_nat_tests.go
```diff
@@ -0,0 +1,713 @@
...
+	if err := DoSetupPeerings(ctx, testCtx.kube, vpcPeerings, externalPeerings, gwPeerings, true); err != nil {
+		return false, nil, fmt.Errorf("setting up NAT gateway peerings: %w", err)
+	}
+
+	if err := WaitReady(ctx, testCtx.kube, testCtx.wrOpts); err != nil {
+		return false, nil, fmt.Errorf("waiting for switches to be ready: %w", err)
+	}
+
+	// Wait for NAT peering to take effect
+	time.Sleep(15 * time.Second)
```
edipascale: just a request if you have time: I'd try to minimize this if possible, I have not observed the need to wait any additional time after `hhfab valb wait` when running tests manually, although of course this might be different for whatever reason. same for the other tests

edipascale (changes requested):
Thanks @pau-hedgehog - it's unfortunate that so much time has passed since the last review that the landscape is completely different, but I think there's some more issues we should address. I'm happy to help tackling them if it means we can merge this faster

pau-hedgehog:
OK, this is passing now after the latest rebases, which includes the removal of the cleanup between tests

pkg/hhfab/rt_nat_tests.go
```diff
@@ -453,49 +630,374 @@ func (testCtx *VPCPeeringTestCtx) gatewayPeeringBidirectionalStaticNATTest(ctx c
...
+					VLAN:   newVLAN,
+					DHCP: vpcapi.VPCDHCP{
+						Enable: true,
+					},
+				},
+			},
+		},
+	}
+	if err := testCtx.kube.Create(ctx, overlapVPC); err != nil {
+		_ = testCtx.kube.Delete(ctx, overlapNS)
```
edipascale: nit: I think you can remove these deletes after a failed create, there should be nothing to delete if you got an error

edipascale (approved):
Thanks for all the iterations! Let's get this in

pkg/hhfab/rt_nat_tests.go
```diff
@@ -0,0 +1,898 @@
...
+		return nil
+	}
+
+	// Test vpc1 -> vpc2 direction (always)
+	if err := testDirection("vpc1->vpc2", vpc1Servers, vpc2Servers, vpc2SubnetStart, vpc2NATPoolStart); err != nil {
+		return err
+	}
+
+	// Test vpc2 -> vpc1 direction (only if vpc1 has static NAT - masquerade doesn't support being a destination)
+	if len(vpc1NATPool) > 0 && !vpc1NATPoolStart.IsUnspecified() {
```
Frostman: Unspecified IP is 0.0.0.0, but I think you're looking for the kinda "empty" one which `!vpc1NATPoolStart.IsValid()`

Frostman (approved):
I think it's generally fine to get merged, but please fix the comments asap in a follow up @pau-hedgehog

PR #987, opened 2025-09-23
ci: fix govulncheck workflow exit handling

pau-hedgehog:
Related to https://github.com/githedgehog/internal/issues/234

### DIFF
--- a/.github/workflows/govulncheck-scan.yaml
+++ b/.github/workflows/govulncheck-scan.yaml
@@ -28,19 +28,14 @@ jobs:
       - name: Run govulncheck
         env:
           CGO_ENABLED: "0"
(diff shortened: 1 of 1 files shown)


edipascale:
not related to the issue here, but we need something similar for the release-test reports, which are only added to the PR as a comment when they pass. Not that the message is particularly informative right now...

pau-hedgehog:
> githedgehog/internal#234

Yes, there are other sub-tasks in https://github.com/githedgehog/internal/issues/234

pau-hedgehog:
> not related to the issue here, but we need something similar for the release-test reports, which are only added to the PR as a comment when they pass. Not that the message is particularly informative right now...

What do you mean exactly?..

PR #988, opened 2025-09-23
release-test: static-external stability

### LINKED ISSUE #932: hlab: static external release test random failure
https://github.com/githedgehog/fabricator/actions/runs/17478219011/job/49643400007
```
01:39:39 ERR FAIL test=StaticExternal error="pinging static external from server-1 in the SE VPC: ping from server-1 to 172.31.255.1: ping failed, expected success"
```

edipascale:
Fix #932 

Note: the route check is not actually needed, I believe, but I'm inclined to keep it anyway - let me know if you think we should remove it (see discussion here for reference: https://github.com/githedgehog/fabricator/issues/932#issuecomment-3324023280)

### DIFF
--- a/pkg/hhfab/release.go
+++ b/pkg/hhfab/release.go
@@ -4,6 +4,7 @@
 package hhfab
 
 import (
+	"bytes"
 	"context"
 	"encoding/xml"
 	"errors"
@@ -270,6 +271,31 @@ func execConfigCmd(hhfabBin, workDir, swName string, cmds ...string) error {
 	return nil
 }
 
+func execShowCmd(hhfabBin, workDir, swName string, cmds ...string) (string, error) {
+	cmd := exec.Command(
+		hhfabBin,
+		"vlab",
+		"ssh",
+		"-n",
+		swName,
+		"-b",
+		"sonic-cli",
+	)
+	for _, c := range cmds {
+		// add escaped double quotes around the command
+		cmd.Args = append(cmd.Args, "-c", fmt.Sprintf("\"%s\"", c))
+	}
+	cmd.Dir = workDir
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	stdout, err := cmd.Output()
+	if err != nil {
+		return "", fmt.Errorf("configuring switch %s: %w, stderr: %s", swName, err, stderr.String())
+	}
+
+	return string(stdout), nil
+}
+
 // Run a command on node nodeName via ssh.
 func execNodeCmd(hhfabBin, workDir, nodeName string, command string) error {
 	cmd := exec.Command(
@@ -353,6 +379,7 @@ func pingFromFabricNode(hhfabBin, workDir, nodeName, ip string, sourceIP string,
 		hhfabBin,
 		"vlab",
 		"ssh",
+		"-b",
 		"-n",
 		nodeName,
 		"ping",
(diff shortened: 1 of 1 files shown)


pkg/hhfab/release.go
```diff
@@ -1339,23 +1366,59 @@ outer:
...
 }
 
+func (testCtx *VPCPeeringTestCtx) checkRouteInSwitch(switchName, route, vrfName string) (bool, error) {
+	cliCmd := fmt.Sprintf("show ip route vrf %s %s", vrfName, route)
+	out, err := execShowCmd(testCtx.hhfabBin, testCtx.workDir, switchName, cliCmd)
+	if err != nil {
+		return false, fmt.Errorf("executing CLI command '%s' on switch %s: %w", cliCmd, switchName, err)
+	}
+
+	return out != "", nil
```
pau-hedgehog, line 1387: should we check for the specific route to be more precise?

pkg/hhfab/release.go
```diff
@@ -1649,8 +1714,18 @@ func (testCtx *VPCPeeringTestCtx) staticExternalTest(ctx context.Context) (bool,
...
 		return nil
 	})
-	slog.Debug("Waiting 1 minute...")
-	time.Sleep(1 * time.Minute)
+	// look for routes in the switch(es) before pinging, see https://github.com/githedgehog/fabricator/issues/932#issuecomment-3322976488
+	if err := testCtx.waitForRoutesInSwitches(ctx, routeCheckSwList, []string{StaticExternalNH, StaticExternalDummyIface}, "VrfV"+vpc1.Name, 3*time.Minute); err != nil {
+		return false, reverts, fmt.Errorf("waiting for routes in switch %s vrf VrfV%s: %w", switchName, vpc1.Name, err)
+	}
+
+	slog.Debug("Pinging from the switch attached to the static external to trigger ARP resolution", "switch", switchName, "vrf", "VrfV"+vpc1.Name, "source-ip", StaticExternalIP, "target", StaticExternalNH)
```
pau-hedgehog, line 1722: Have you analyzed if we can mask any potential issue by doing this? It's a similar case than the warm-up pings in the connectivity checks

pau-hedgehog (approved):
Very minor comments. I'm testing this and has passed several times in a row in env-1

pkg/hhfab/release.go
edipascale: it's not trivial to do so in the generic case; if your query is `show ip route 10.10.2.3/32`, what you get depends on what is the longer prefix matching route in the routing table, so for example you could get:
* 10.10.2.3/32
* 10.10.2.0/24
* 10.0.0.0/8
* 0.0.0.0/0
* ... and any other possible IP prefix which includes that route and was advertised in that VRF

but also, if you get anything at all on stdout, it's because there was a route. The only annoying thing is that with the current infra we combine stderr and stdout, so we cannot check that what we get was actually a stderr error message, but once we transition [to the native ssh library](https://github.com/githedgehog/fabricator/pull/1006) we could do that. The alternative right now could be to check for the routing flag states that are printed at the beginning of `show ip route` but I figured it wasn't worth it

pkg/hhfab/release.go
pau-hedgehog: Sounds good for now

PR #990, opened 2025-09-24
fix(release): always use single resulting port breakout

Frostman:
It allows to avoid issues with the pipelines and any other potential issues with finding the right port for the test. We just need to test the fact that we can change a breakout, doesn't matter which mode we're going to use.

Fixes githedgehog/internal#221


edipascale (approved):
LGTM, only doubt I have is if all switches have such a non-default, single offset breakout mode, else the test might fail

pau-hedgehog (approved):

Frostman:
Yeah, I've looked at the switch catalog and seems like we always have a few options for 1xXYZ as it would be some 1x40G or 1x25G and etc

PR #1014, opened 2025-10-01
fix(ci): always publish release-test results

edipascale:
currently they are only published on success, which is pretty useless. also print individual failures for each run of the same test

part of https://github.com/githedgehog/internal/issues/220


edipascale:
finally after many retries we got a failure, output looks good:
image

Issue #1015, opened 2025-10-01
SONiC installer always fails on a first installation attempt


Frostman:
As a bug, it's very minor as it always succeeded on the following attempts, but it slows down VLAB quite a lot as the switch goes twice through the installer, and it could be masking some other issues. IIRC it only happens on the virtual switch, but worse checking the hardware ones as well.

pau-hedgehog:
shall I implement also https://github.com/githedgehog/internal/issues/219 so we have a measure about vs installation duration? (maybe just as a debug log)

PR #1019, opened 2025-10-02
fix(onie): resize image partition table

pau-hedgehog:
Fixes: #1015

### DIFF
--- a/pkg/fab/README.md
+++ b/pkg/fab/README.md
@@ -109,11 +109,31 @@ oras push "ghcr.io/githedgehog/fabricator/k9s:${K9S_VERSION}" k9s
 
 ## ONIE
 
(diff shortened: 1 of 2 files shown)


pau-hedgehog:
ONIE installer fails when the GPT partition table doesn't reflect the actual disk size. The original image is resized at runtime, but the GPT metadata still indicates 1GB of usable space. This causes the first install to create an undersized partition and fail. After a retry loop, the second attempt succeeds with the correct partition.

@Frostman, could we have a onie image processed as in README and pushed?

I've tested processing the image from my VLAB cache and worked like a charm when running `vlab up -f` again

Frostman:
@pau-hedgehog I've uploaded following your instructions and added bump for versions.go to your commit so we can test in this PR

Frostman:
I confirm it's working on my vlab

PR #1022, opened 2025-10-03
chore: validate dummyIP clash in fabNode

### LINKED ISSUE #1016: Replicate all hydration-level checks to happen in Fabricator object validation too
We need to copy all the checks we have in the `/fabricator/pkg/hhfab/hydrate.go` to the Fabricator's CRD validation (while keeping hydrate as is), such as:
- dummy interface IP on the FabNode

edipascale:
Fix #1016


api/fabricator/v1beta1/fabnode_types.go
```diff
@@ -100,6 +103,36 @@ func (n *FabNode) Validate(_ context.Context, fabCfg *FabConfig, allowNotHydrate
...
+					dummyIPs[ip.Addr()] = true
+				}
+			}
+
+			if _, exists := dummyIPs[dummyAddr.Addr()]; exists {
+				return fmt.Errorf("dummy IP %s already in use", dummyAddr.String()) //nolint:goerr113
+			}
+		}
+
 		managementAddr, err := n.Spec.Management.IP.Parse()
```
Frostman, line 136: Could you please do the same for the management IPs?

edipascale:
this is still not working in that I can duplicate a gateway fabnode without the validation complaining, but I have to drop for today, I'll continue on Monday

PR #1051, opened 2025-10-20
chore(vlab): add lgtm observability node

pau-hedgehog:


pau-hedgehog:
You can bring the VLAB+LGTM node locally with:

```
just lint && just oci=http push && bin/hhfab init --registry-repo localhost:30000 --dev --obs --gw -f && bin/hhfab vlab gen --eslag-leaf-groups=2 --spines-count=2 --bundled-servers=0 --eslag-serv
ers=1 --unbundled-servers=1 && bin/hhfab vlab up -f -v --controls-restricted=false
```
Once VLAB is ready you can get to Grafana on host_ip:3000 and you will have logs/metrics from the VLAB readily available

`--controls-restricted=false` is required because I was having issues pushing some images to the registry. Some of them are directly downloaded for now

pkg/support/kuberesources.go
```diff
@@ -153,7 +153,7 @@ func redactAlloyTargets(targets *alloy.Targets) {
 }
 
 func redactAlloyTarget(target *alloy.Target) {
-	if target.BasicAuth.Password != "" {
+	if target.BasicAuth != nil && target.BasicAuth.Password != "" {
```
Frostman: Could you please make a separate PR with it so we can merge it

pkg/fab/comp/k3s/localpathconfig.go
```diff
@@ -0,0 +1,201 @@
+// Copyright 2025 Hedgehog
```
Frostman: I wasn't digging into what's done here, why is it required?

PR #1161, opened 2025-11-16
chore(vlab): fix server vm dhcp mtu

pau-hedgehog:


pau-hedgehog:
> I see a bit of a problem with the approach as we sometimes need to test nativeVLAN which means attachments w/o VLANs

There's yet another problem with the approach. Jumbo frames crash the virtual switch due to a known limitation.

Let me see if I can do something about those 2 issues

PR #1172, opened 2025-11-19
New DHCP Release Tests: Static Lease and Pool Depletion

pau-hedgehog:


pkg/hhfab/rt_single_vpc_suite.go
```diff
@@ -1006,6 +1017,622 @@ func (testCtx *VPCPeeringTestCtx) dhcpRenewalTest(ctx context.Context) (bool, []
...
+			return fmt.Errorf("getting IP address from %s: %w", testServer.Name, err)
+		}
+
+		assignedIP := strings.TrimSpace(ipOut)
+		if assignedIP != staticIP.String() {
+			return fmt.Errorf("server %s did not get static IP: expected %s, got %s", testServer.Name, staticIP.String(), assignedIP) //nolint:goerr113
+		}
+
+		slog.Info("Static IP assignment verified", "description", description, "server", testServer.Name, "ip", assignedIP)
+
```
edipascale: not sure how feasible it would be, but if you use the first IP in the dynamic range for your static IP test and then try to request an address using DHCP, we could double check that the static assignment is skipped as expected. just an idea - doesn't have to be implemented with this PR even if we think it's doable

pkg/hhfab/rt_single_vpc_suite.go
```diff
@@ -1006,6 +1017,622 @@ func (testCtx *VPCPeeringTestCtx) dhcpRenewalTest(ctx context.Context) (bool, []
...
+	// Group attachments by VPC and subnet
+	type vpcSubnetKey struct {
+		vpcName    string
+		subnetName string
+	}
+
+	serversByVPCSubnet := make(map[vpcSubnetKey][]ServerWithInterface)
+
+	for _, attach := range vpcAttaches.Items {
```
edipascale: this code is duplicated from the previous test and quite bulky, it's maybe worth to put it into a helper function instead

edipascale (changes requested):
thanks for adding this extra coverage @pau-hedgehog! I left some comments mainly aimed at reducing code duplication and making it easier to read and maintain in the future, but the tests themselves look good to me

PR #1174, opened 2025-11-19
chore(ci): publish release tests results for custom vlab workflows

pau-hedgehog:


.github/workflows/custom-vlab.yaml
```diff
@@ -85,3 +85,26 @@ jobs:
...
       debug: ${{ inputs.pause_on_failure }}
       pause_on_failure: ${{ inputs.pause_on_failure }}
+
+  publish-test-results:
+    if: ${{ inputs.releasetest && !cancelled() }}
+    runs-on: lab
+    needs:
+      - vlab
+
+    permissions:
```
Frostman: Custom VLAB is manually triggered so it shouldn't write to any branch checks or PRs, you may want to just write to the workflow summary instead

PR #1189, opened 2025-11-23
Diagram: add drawio VPC layer

pau-hedgehog:
Generates layer with VPC information:
image


Frostman (approved):

pau-hedgehog:
> @pau-hedgehog could you please add a cli flag to the diagram PR to enable/disable showing VPCs on the diagram? I think we can keep it enabled by default, but would be nice to be able to turn it off to have more "clean" topo overview

We already have several options:

- Without `--live` there is no additional layer information other that the ones we can get from the YAMLs
- If we generate a diagram with `--live`, we can toogle layers on draw.io:
image

Is this enough or would you still like to be able to decide which layers do we generate whith `--live`? Mind that after vpc and link speed I had in mind to work eventually on underlay, etc

PR #1196, opened 2025-11-26
fix(ci): fix upgrade skip logic (again)

edipascale:


edipascale:
I've removed the label to make sure upgrade jobs were running without it, and they do; the PR appears to do its job just fine

PR #1197, opened 2025-11-26
Diagram: add drawio link speed layer in drawio

pau-hedgehog:
Adds new layer in draw.io with live operational status (green/UP red/DOWN). Adds NOS interface names as tooltips to the API interface label:
image

### DIFF
--- a/pkg/hhfab/diagram/drawio.go
+++ b/pkg/hhfab/diagram/drawio.go
@@ -31,7 +31,8 @@ type MxGraphModel struct {
 }
 
 type Root struct {
-	MxCell []MxCell `xml:"mxCell"`
+	MxCell     []MxCell     `xml:"mxCell"`
+	UserObject []UserObject `xml:"UserObject"`
 }
 
 type MxCell struct {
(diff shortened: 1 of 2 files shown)


pkg/hhfab/diagram/drawio.go
```diff
@@ -1785,3 +1846,154 @@ func createVPCLegend(model *MxGraphModel, vpcs map[string]*VPCInfo, serverBottom
...
+		// 2. Create Ethernet interface name labels (source and target) with status-based colors
+		srcNOS := edgeData.Link.Properties[PropSourcePortNOS]
+		tgtNOS := edgeData.Link.Properties[PropTargetPortNOS]
+		srcStatus := edgeData.Link.Properties[PropSourcePortStatus]
+		tgtStatus := edgeData.Link.Properties[PropTargetPortStatus]
+
+		// Source Ethernet label (if available)
+		if srcNOS != "" {
+			// Shorten "Ethernet" to "Eth" for display
+			displaySrcNOS := strings.Replace(srcNOS, "Ethernet", "Eth", 1)
```
Frostman: Ports should be named exactly as we have them named in the API. You can add an alt name on hover or in parentheses

Frostman (changes requested):
Didn't look for the rest of the changes, but it's very important everything user-facing to have port names from our API.

pkg/hhfab/diagram/drawio.go
pau-hedgehog: So this is a separate layer with L1 information overlaid on top of the base layer. I thought for troubleshooting purposes it made sense to have it like this but it is true that it does not contain the front panel port ID so it beats my original idea. The base layer has still the ports with the API names. 

I will try to implement the hover idea on draw.io

Frostman (dismissed):
@pau-hedgehog all looks ok, but please rebase

PR #1215, opened 2025-12-05
chore: no more need to have anonymous probe for zot

Frostman:


pau-hedgehog:
Is this stale @mrbojangles3 ? Should we close?

Issue #1232, opened 2025-12-11
iperf error: client: running command: cancelled: context canceled: , server: running command: Process exited with status 124:


pau-hedgehog:
```
07:04:48 ERR FAIL test="Multi-Subnets no restrictions" error="iperf server-03 -> server-01: client: running command: cancelled: context canceled: , server: running command: Process exited with status 124: "
```
https://github.com/githedgehog/fabricator/actions/runs/20123749606/job/57749485615

PR #1242, opened 2025-12-14
chore(testing): add retries to iperf via ss check on server

pau-hedgehog:
Fixes https://github.com/githedgehog/fabricator/issues/1232


edipascale (commented):
thanks @pau-hedgehog, just a couple of questions. Also please squash the code related commits (i.e. everything but the gitignore) into a single one if possible, as it's really iterative fixes over the same function

edipascale (approved):

PR #1265, opened 2025-12-16
Capture basic show-tech from switches via console

pau-hedgehog:


Frostman:
@pau-hedgehog why replace ssh with console?

pkg/hhfab/vlabhelpers.go
```diff
@@ -582,6 +628,40 @@ func (c *Config) VLABShowTech(ctx context.Context, vlab *VLAB) error {
 		}
 	}()
 
+	var switchConsoleScriptPath string
+	var switchConsoleCleanup func()
+	var serverConsoleScriptPath string
+	var serverConsoleCleanup func()
+	if _, err := exec.LookPath(VLABCmdExpect); err == nil {
+		if cleanup, path, err := c.prepareShowTechConsoleScript(); err == nil {
+			switchConsoleCleanup = cleanup
```
Frostman: you can just defer cleanup here (nit pick, doesn't have to fix)

pkg/hhfab/vlabhelpers.go
```diff
@@ -582,6 +628,40 @@ func (c *Config) VLABShowTech(ctx context.Context, vlab *VLAB) error {
...
+			switchConsoleScriptPath = path
+		} else {
+			slog.Warn("Failed to prepare console show-tech script; console fallback disabled", "err", err)
+		}
+		if cleanup, path, err := c.prepareShowTechServerConsoleScript(); err == nil {
+			serverConsoleCleanup = cleanup
```
Frostman: same here (nit pick, doesn't have to fix)

cmd/hhfab/main.go
```diff
@@ -423,6 +423,19 @@ func Run(ctx context.Context) error {
 		},
 	}
 
+	showTechConsoleFlags := []cli.Flag{
+		&cli.StringFlag{
+			Name:    "switch-username",
+			Usage:   "switch username for console show-tech",
+			EnvVars: []string{hhfab.VLABEnvSwitchUser, "HHFAB_SWITCH_USERNAME"},
```
Frostman: One option for env variable is enough

pau-hedgehog:
> Need to get rid of the secret inherit and few minor comments, let me know when you're ready, I'll add secrets into runners

Thanks for looking into this. I've addressed the comments. Let's re-trigger when runners get the secrets

pau-hedgehog:
@Frostman this is ready to add secrets into runners so we can merge and hopefully start getting more insights on the failures

Frostman (dismissed):

PR #1117, opened 2025-11-09
chore(show-tech): gather diagnostics for physical switches

pau-hedgehog:

PR #1274, opened 2025-12-18
chore(show-tech): fix hardware switch collection

pau-hedgehog:
Comes from https://github.com/githedgehog/fabricator/pull/1117/files but just fixing the listing of hardware switches. I will fix the collection when VLAB is being shut down in another PR

### DIFF
--- a/pkg/hhfab/vlabhelpers.go
+++ b/pkg/hhfab/vlabhelpers.go
@@ -598,7 +598,12 @@ func (c *Config) VLABShowTech(ctx context.Context, vlab *VLAB) error {
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


pau-hedgehog:
Note to @Frostman, I couldn't figure out how to make the collection work when there is some failure and the HLAB contexts close. But  this will give us some additional information, specially when we merge https://github.com/githedgehog/fabricator/pull/1289
I split it but we could merge https://github.com/githedgehog/fabricator/pull/1289 to https://github.com/githedgehog/fabricator/pull/1274 and then to master. As you prefer

Issue #1223, opened 2025-12-09
H/VLAB: Error inspecting BGP -- neighbor not established


pau-hedgehog:
```
23:05:01 ERR Reported  err="switch leaf-02: vrf default: neighbor \"172.30.8.0\" is not established"
23:05:01 ERR Inspecting BGP reveals some errors err="inspect function reported 1 
```
https://github.com/githedgehog/fabricator/actions/runs/20045282024/job/57489765171

pau-hedgehog:
Two new hits:
     - https://github.com/githedgehog/fabricator/actions/runs/20828463453/job/59836910879 (pau/show-tech-for-each-failed-relase-test: v-up25.05-mesh-gw-iso-l2vni / (25.05) Run and test VLAB)
     - https://github.com/githedgehog/fabric/actions/runs/20827883119/job/59834364769 (master: v-up25.05-mesh-iso-l2vni / (25.05) Run and test VLAB)

pau-hedgehog:
It seems even with the retries we are getting failures:
```
11:59:11 INF BGP inspect retry attempt number=2
11:59:21 INF BGP inspect retry attempt number=3
11:59:31 ERR BGP inspection has errors after all retries errors="[switch ds4000-01: vrf default: neighbor \"172.30.128.17\" is not established switch ds4000-02: vrf default: neighbor \"172.30.128.35\" is not established]"
11:59:31 ERR Failed to inspect fabric took=2m31.344396327s
```
https://github.com/githedgehog/fabricator/actions/runs/22177620949/attempts/1

pau-hedgehog:
So in the case of HLAB I realized the neighbors not being established were the Gateway ones:
```
  11:59:31 ERR BGP inspection has errors after all retries errors="[switch ds4000-01: vrf default: neighbor \"172.30.128.17\" is not established switch ds4000-02: vrf default: neighbor \"172.30.128.35\" is not established]"      
```
<img width="1384" height="874" alt="Image" src="https://github.com/user-attachments/assets/7c453893-0403-4dd8-b92e-cd186cb67b49" />

I'm afraid we may be having issues again between FRR and dataplane, @Fredi-raspall @Frostman :

Gateway BGP showing neighbors never established (show-tech-output/gateway-1-show-tech.log):
```
=== Executing: vtysh -c 'show bgp summary' ===

IPv4 Unicast Summary:
BGP router identifier 172.30.8.6, local AS number 65534 VRF default vrf-id 0
[...]

Issue #981, opened 2025-09-19
Still icmp VLAB errors


pau-hedgehog:
https://github.com/githedgehog/fabricator/actions/runs/17856635923/job/50777467047
```
2025-09-19T11:51:46.0902180Z 11:51:46 ERR Test connectivity failed ping=6 iperf=0 curl=0 took=51.693375537s errors="ping server-09 -> server-07: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-08 -> server-07: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-06 -> server-04: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-07 -> server-09: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-07 -> server-08: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-04 -> server-06: should be reachable but ping failed (sent 5, rcvd 0, expected true)"
2025-09-19T11:51:46.0923737Z 11:51:46 ERR FAIL test="Multi-Subnets no restrictions" error="ping server-09 -> server-07: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-08 -> server-07: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-06 -> server-04: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-07 -> server-09: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-07 -> server-08: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-04 -> server-06: should be reachable but ping failed (sent 5, rcvd 0, expected true)"
2025-09-19T11:51:46.0927366Z 11:51:46 INF SKIP test="Multi-Subnets isolation" reason="There are virtual switches"
2025-09-19T11:51:46.0928012Z 11:51:46 INF SKIP test="Multi-Subnets with filtering" reason="There are virtual switches"
2025-09-19T11:51:46.0928521Z 11:51:46 INF SKIP test=StaticExternal reason="There are virtual switches"
2025-09-19T11:51:46.0929150Z 11:51:46 INF ** Finished test suite suite="Multi-Subnets VPC Suite" duration=2m2s
2025-09-19T11:51:46.0930044Z 11:51:46 INF Test suite results suite="Multi-Subnets VPC Suite"
[...]

PR #1284, opened 2025-12-22
Address CI VLAB stability with targeted waits when using Virtual Switches

pau-hedgehog:
This PR contains a few commits that were tested in a previous experimental PR that was addressing CI failures incrementally and grew out of control: https://github.com/githedgehog/fabricator/pull/1201

Addresses: 
https://github.com/githedgehog/fabricator/issues/981 https://github.com/githedgehog/fabricator/issues/1223

### DIFF
--- a/pkg/hhfab/rt_base.go
+++ b/pkg/hhfab/rt_base.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	waitAppliedFor = 15 * time.Second
-	waitTimeout    = 5 * time.Minute
+	waitAppliedFor   = 15 * time.Second
+	waitAppliedForVS = 60 * time.Second // extended wait for virtual switches
+	waitTimeout      = 5 * time.Minute
 )
 
 var (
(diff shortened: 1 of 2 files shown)


edipascale (changes requested):
I'm not convinced by this change:
* could you add some context in the PR description on why is it needed? was it an empirical check that made you realize virtual switches need more time? Is there something more specific that we could be polling / checking instead? Apologies if you have already discarded all these options, but I have no visibility on this.
* If we really need additional time for virtual switches, can we not just increase the `AppliedFor` timer in that case? What's the purpose of adding a separate stabilization period?
* (nit) In release tests we already check whether there are virtual switches so that we can set the related skip flag; you could compute that information once in the caller and use it for both

pau-hedgehog:
> I'm not convinced by this change:
> 
> * could you add some context in the PR description on why is it needed? was it an empirical check that made you realize virtual switches need more time? Is there something more specific that we could be polling / checking instead? Apologies if you have already discarded all these options, but I have no visibility on this.
> * If we really need additional time for virtual switches, can we not just increase the `AppliedFor` timer in that case? What's the purpose of adding a separate stabilization period?
> * (nit) In release tests we already check whether there are virtual switches so that we can set the related skip flag; you could compute that information once in the caller and use it for both

I followed your suggestion and simplified with `AppliedFor`

Issue #1239, opened 2025-12-13
Context deadline exceeded after no switches becoming ready


pau-hedgehog:
https://github.com/githedgehog/dataplane/actions/runs/20069038523/job/57568210055
```
16:39:52 INF Switches status notReady="[leaf-01 leaf-02 leaf-03 leaf-04 leaf-05 spine-01 spine-02]" notUpdated=[]
16:40:22 INF Switches status notReady="[leaf-01 leaf-02 leaf-03 leaf-04 leaf-05 spine-01 spine-02]" notUpdated=[]
16:40:52 INF Switches status notReady="[leaf-01 leaf-02 leaf-03 leaf-04 leaf-05 spine-01 spine-02]" notUpdated=[]
16:41:22 INF Switches status notReady="[leaf-01 leaf-02 leaf-03 leaf-04 leaf-05 spine-01 spine-02]" notUpdated=[]
16:41:52 INF Switches status notReady="[leaf-01 leaf-02 leaf-03 leaf-04 leaf-05 spine-01 spine-02]" notUpdated=[]
16:41:52 INF Switches status notReady="[leaf-01 leaf-02 leaf-03 leaf-04 leaf-05 spine-01 spine-02]" notUpdated=[]
16:42:16 ERR Failed to wait for ready err="waiting for ready: context deadline exceeded"
16:42:16 INF Retry attempt number=2
16:42:16 WRN Failed to inspect err="context cancelled during retry: context deadline exceeded"
16:42:32 DBG Show tech collected successfully entry=server-05 output=/home/runner/_work/dataplane/dataplane/show-tech-output/server-05-show-tech.log
16:42:32 DBG Show tech collected successfully entry=server-04 output=/home/runner/_work/dataplane/dataplane/show-tech-output/server-04-show-tech.log
16:42:32 DBG Show tech collected successfully entry=server-08 output=/home/runner/_work/dataplane/dataplane/show-tech-output/server-08-show-tech.log
16:42:32 DBG Show tech collected successfully entry=server-02 output=/home/runner/_work/dataplane/dataplane/show-tech-output/server-02-show-tech.log
16:42:32 DBG Show tech collected successfully entry=server-10 output=/home/runner/_work/dataplane/dataplane/show-tech-output/server-10-show-tech.log
16:42:32 DBG Show tech collected successfully entry=server-07 output=/home/runner/_work/dataplane/dataplane/show-tech-output/server-07-show-tech.log
[...]

PR #1295, opened 2026-01-08
ci: increase upgrade job timeout

pau-hedgehog:
This was used in a previous branch to diagnose upgrade job failures as early job cancellation was preventing the gathering of show-tech

We need it again (at least temporarily) to properly investigate issue https://github.com/githedgehog/fabricator/issues/1239 , because we can hit it during the upgrade step and the job will time out, leaving us without proper diagnostic data


edipascale (approved):
thanks for the explanation @pau-hedgehog, lgtm

Frostman (approved):

PR #1324, opened 2026-01-16
set gw default log level via hhfab vlab gen argument

### LINKED ISSUE #1322: support setting gateway logging config as flags in hhfab vlab gen

edipascale:
Fix #1322

### DIFF
--- a/.github/workflows/run-vlab.yaml
+++ b/.github/workflows/run-vlab.yaml
@@ -87,13 +87,18 @@ on:
         type: string
         required: false
         default: ""
+      custom_gen_args:
+        description: "Custom arguments to be passed to hhfab vlab gen"
+        type: string
+        required: false
+        default: ""
 
 permissions:
   contents: read
 
 env:
   # global workflow configs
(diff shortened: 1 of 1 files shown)


api/fabricator/v1beta1/fabricator_types.go
```diff
@@ -311,6 +311,7 @@ type GatewayConfig struct {
 	Observability *GatewayObservability `json:"observability,omitempty"`
 	DepAgentless  bool                  `json:"agentless,omitempty"`
 	Communities   map[string]string     `json:"communities,omitempty"`
+	LogLevel      string                `json:"loglevel,omitempty"`
```
Frostman: We shouldn't be adding it here, it's already exposed on the Gateway object

Frostman (changes requested):

api/fabricator/v1beta1/fabricator_types.go
Frostman: Yes, info needs to be stored somehow, but Fabricator is a wrong place for it as it's a runtime config that user can change and controller is supposed to apply it, in that case it'll somewhat an override of what's already defined in the gateway object.

PR #1378, opened 2026-01-27
chore(test): add DHCP static lease test

pau-hedgehog:
Comes from https://github.com/githedgehog/fabricator/pull/1172

Includes only the Static DHCP lease test which is not adding any flakiness


edipascale (dismissed):
Thanks @pau-hedgehog, this looks much more streamlined to me. I've left a few minor comments, they could be addressed in a separate PR, but if not please squash commits into a single one

PR #1402, opened 2026-02-02
chore(testing): fix iperf json parsing

pau-hedgehog:
Fixes https://github.com/githedgehog/fabricator/issues/1401

### DIFF
--- a/pkg/hhfab/testing.go
+++ b/pkg/hhfab/testing.go
@@ -4,6 +4,7 @@
 package hhfab
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


edipascale (dismissed):
LGTM, just a question: were you able to check that it works? Specifically I'm worried about errors at the end of the json output interfering with the parsing, or does the unmarshal function stop when it reaches the end of its block?

pau-hedgehog:
> LGTM, just a question: were you able to check that it works? Specifically I'm worried about errors at the end of the json output interfering with the parsing, or does the unmarshal function stop when it reaches the end of its block?

Sorry, this was supposed to be a draft before having an AI review first and try to verify it on CI

PR #1407, opened 2026-02-03
Add demo CI workflow

pau-hedgehog:
Related to https://github.com/githedgehog/docs/issues/268

### DIFF
--- a/.github/workflows/demo-vlab.yaml
+++ b/.github/workflows/demo-vlab.yaml
@@ -1,0 +1,228 @@
+name: Demo VLAB
+
+on:
+  push:
+    paths:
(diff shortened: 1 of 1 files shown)


pau-hedgehog:
I've tested this here because it is where we have the vlab runner access but it might make more sense to have in the docs repo. Se discussion in https://github.com/githedgehog/internal/issues/311

Frostman (commented):
Just generic thoughts - it'll be VERY difficult to maintain it in sync with the docs, should we instead support running some test script, have it in the docs repo and embed parts of it in the docs instead? Not sure how easy could it be done, just thoughts.

Frostman:
Closed as stale, not needed atm

Issue #1292, opened 2026-01-07
HLAB: iPerf completed but throughput below expected threshold on


pau-hedgehog:
1 job - Affects hardware lab jobs (h-*) ONLY

  1. https://github.com/githedgehog/fabricator/actions/runs/20739829522/job/59544618554
    - Branch: master
    - Job: h-gw-iso-l2vni-rt / Run and test hybrid VLAB

  Raw ERR Lines

```
2026-01-06T08:02:40.8784121Z 08:02:40 ERR Test connectivity failed ping=0 iperf=1 curl=0 took=3m59.550324027s errors="iperf server-7 -> server-1: client: iperf3 send speed too low, server: running command: cancelled: context canceled:  (sent 7400.60 Mbps, rcvd 7371.77 Mbps, min expected 8200.00 Mbps)"
2026-01-06T08:02:40.8789745Z 08:02:40 ERR FAIL test="RoCE flag and basic traffic marking" error="testing connectivity with DSCP opts: iperf server-7 -> server-1: client: iperf3 send speed too low, server: running command: cancelled: context canceled:  (sent 7400.60 Mbps, rcvd 7371.77 Mbps, min expected 8200.00 Mbps)"
2026-01-06T08:02:40.8808013Z 08:02:40 ERR FAIL test="RoCE flag and basic traffic marking" error="testing connectivity with DSCP opts: iperf server-7 -> server-1: client: iperf3 send speed too low, server: running command: cancelled: context canceled:  (sent 7400.60 Mbps, rcvd 7371.77 Mbps, min expected 8200.00 Mbps)"
```

PR #1446, opened 2026-02-11
chore(testing): add retries for iperf too slow failure

pau-hedgehog:
Fixes https://github.com/githedgehog/fabricator/issues/1292


pau-hedgehog:
> LGTM; as a nit we might want to rename the SpeedTooLow flag to indicate that we are only setting it when the send speed is too low, else it can be a little confusing in the future when we forget why it was added in the first place; or alternatively split it into two flags (one for send speed and one for receive speed) and only retry if the send flag is set

Changed to `SendSpeedTooLow`

Issue #1234, opened 2025-12-11
gateway K8s node does not become ready after flatcar upgrade


pau-hedgehog:
```
20:41:31 INF upgrade(control-1): Dec 11 20:41:30.691 INF Upgrading Flatcar
20:41:31 INF upgrade(control-1): Dec 11 20:41:30.691 INF Upgrading Flatcar to version=v4459.2.1
20:41:33 INF upgrade(control-1): Dec 11 20:41:32.290 DBG flatcar-update: Forcing update...
20:41:33 INF upgrade(control-1): Dec 11 20:41:32.298 DBG flatcar-update: 2025/12/11 20:41:32 socat[28702] E read(5, 0x557a44cf8000, 8192): Connection reset by peer
20:41:37 INF upgrade(control-1): Dec 11 20:41:36.282 DBG flatcar-update: 2025/12/11 20:41:36 socat[28902] E read(5, 0x557a44cf8000, 8192): Connection reset by peer
20:41:51 INF upgrade(control-1): Dec 11 20:41:50.096 DBG flatcar-update: 2025/12/11 20:41:50 socat[28907] E read(5, 0x55e3d4dfa000, 8192): Connection reset by peer
20:41:51 INF upgrade(control-1): Dec 11 20:41:50.395 DBG flatcar-update: 2025/12/11 20:41:50 socat[28974] E read(5, 0x557a44cf8000, 8192): Connection reset by peer
20:41:54 INF upgrade(control-1): Dec 11 20:41:52.976 DBG flatcar-update: 2025/12/11 20:41:52 socat[29065] E read(5, 0x557a44cf8000, 8192): Connection reset by peer
20:41:58 INF upgrade(control-1): Dec 11 20:41:57.939 DBG flatcar-update: Done, please make sure to reboot either manually or through your reboot manager (check locksmithd/FLUO)
20:41:58 INF upgrade(control-1): Dec 11 20:41:57.943 DBG flatcar-update: 2025/12/11 20:41:57 socat[28700] W exiting on signal 15
20:41:58 INF upgrade(control-1): Dec 11 20:41:57.944 DBG flatcar-update: 2025/12/11 20:41:57 socat[28697] W exiting on signal 15
20:41:58 INF upgrade(control-1): Dec 11 20:41:57.945 INF Flatcar upgrade completed
20:41:58 INF upgrade(control-1): Dec 11 20:41:57.945 WRN A reboot is necessary for the changes to take effect
[...]

PR #1455, opened 2026-02-13
chore(vlab): gather show-tech after k8s node timeout

pau-hedgehog:
Related to https://github.com/githedgehog/fabricator/issues/1234

### DIFF
--- a/pkg/hhfab/vlabrunner.go
+++ b/pkg/hhfab/vlabrunner.go
@@ -567,6 +567,8 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 				case <-ctx.Done():
 					slog.Error("Failed to wait for k8s api", "err", readyErr)
 
+					c.CollectVLABDebug(ctx, vlab, opts)
+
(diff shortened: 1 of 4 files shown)


pkg/hhfab/vlabrunner.go
```diff
@@ -567,6 +567,11 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 				case <-ctx.Done():
 					slog.Error("Failed to wait for k8s api", "err", readyErr)
 
+					// Use fresh context since parent ctx is cancelled
+					debugCtx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+					defer cancel()
```
Frostman: it'll defer it within the `VLABRun` which is bad as you run it in the retry loop, you should have some func called and that func should defer within itself, maybe you want to extend the CollectVLABDebug for it

Frostman (changes requested):
Do you think it has a chance to work? If ctx is cancelled it'll kill VM right away AFAIK. Have you tested it?

PR #1456, opened 2026-02-13
chore(just): build golangci-lint with right go version

pau-hedgehog:
Not sure how you handle this on your development environment but each time we bump go I have issues with golangci-lint

trying to address this once and for all


Frostman (changes requested):
It's not going to help as it'll not be even called as golangci-lint binary will be already there, you need to add golang version somehow to the name of the golangci-lint binary which aren't going to be very nice too IMO. It's happening vry rarely and removing bin folder shouldn't be a problem like every half year or so

PR #1463, opened 2026-02-16
ci: change CI workflow dispatch behavior

pau-hedgehog:
This tries to be a DX improvement.

Adds toogle for release tests and hlab, this way it doesn't run hlab unless requested: image

Instead of:
image

I am fail-proofing some branches to reduce CI flakiness and having hlab run by default is slowing down everything

push and PR triggers are unaffected

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,14 +26,14 @@ on:
     - cron: "0 10 * * *" # ~2am PST
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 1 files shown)


pau-hedgehog:
push:
- [ ]  no labels
- [ ] +hlab
- [ ] -vlab
- [ ] -upgrades

workflow dispatch:
- [ ] no options
- [ ] release tests
- [ ] hlab

PR #1470, opened 2026-02-19
chore(test): add static external release test

pau-hedgehog:


edipascale (changes requested):
Thanks @pau-hedgehog, just a few small changes please

as a side note we might want to add external peerings towards the static external in other tests instead of / as well as in the dedicated test you added, but I'm fine either way, your approach works too

Frostman (dismissed):

Issue #1483, opened 2026-02-24
HLAB: Gateways status notReady


pau-hedgehog:
https://github.com/githedgehog/gateway/actions/runs/22220090113/job/64273399787
```
14:35:52 INF Inspecting fabric
14:35:52 INF Waiting for fabricator ready timeout=30m0s
14:35:52 INF Fabricator ready gen=1 version=v0.43.1-237-gc441bdd1-dirty-KS1353
14:35:52 INF Waiting for switches and gateways ready appliedFor=30s timeout=30m0s
14:35:52 INF Expected switches agent=v0.106.0 switches="[ds3000-01 ds3000-02 ds3000-03 ds4000-01 ds4000-02 sse-c4632-01]"
14:35:52 INF Expected gateways dataplane=v0.12.0 gateways=[gateway-1]
14:35:52 INF Switches status notReady="[ds3000-01 ds3000-02 ds3000-03 sse-c4632-01]" notUpdated=[]
14:35:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:36:22 INF Switches status notReady="[ds3000-01 ds3000-02 ds3000-03 sse-c4632-01]" notUpdated=[]
14:36:22 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:36:52 INF Switches status notReady="[ds3000-01 ds3000-02 ds3000-03 sse-c4632-01]" notUpdated=[]
14:36:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:37:22 INF Switches status notReady="[ds3000-02 ds3000-03 sse-c4632-01]" notUpdated=[]
14:37:22 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:37:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:38:22 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:38:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:39:22 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:39:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:40:22 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:40:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:41:22 INF Gateways status notReady=[gateway-1] notUpdated=[]
14:41:52 INF Gateways status notReady=[gateway-1] notUpdated=[]
[...]

Issue #1468, opened 2026-02-17
Decouple VLAB process handling and cleanup for easier debug collection


Frostman:

pau-hedgehog:
Another one where we cannot diagnose due to failed show-tech: https://github.com/githedgehog/fabricator/issues/1487

PR #1484, opened 2026-02-24
ci: increase timeout for hlab jobs without release test

pau-hedgehog:
Related to https://github.com/githedgehog/fabricator/issues/1483

### DIFF
--- a/.github/workflows/run-vlab.yaml
+++ b/.github/workflows/run-vlab.yaml
@@ -102,7 +102,7 @@ jobs:
   run:
     if: ${{ !inputs.skip }}
     runs-on: ${{ inputs.hybrid && 'hlab' || 'vlab' }}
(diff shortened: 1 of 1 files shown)


Frostman:
@pau-hedgehog gateway taking that much time looks like some problem with the gw itself

pau-hedgehog:
> @pau-hedgehog gateway taking that much time looks like some problem with the gw itself

Yes, but without diagnostics we cannot say. And catching this live in CI and investigating is virtually impossible

pau-hedgehog:
> @pau-hedgehog gateway taking that much time looks like some problem with the gw itself

@Frostman, the thing is that either we have hhfab and CI timeouts aligned or we must work on https://github.com/githedgehog/fabricator/issues/1468 as a priority to get proper diagnostics and investigate before release. Otherwise we are assuming risks on the coming release

Issue #1254, opened 2025-12-15
VLAB/HLAB: some switches status never gets ready


pau-hedgehog:
https://github.com/githedgehog/fabricator/actions/runs/19859236571/job/56905389997
```
15:17:31 INF Gateways status notReady=[gateway-1] notUpdated=[]
15:18:01 INF Switches status notReady="[sse-c4632-01 ds3000-01 ds3000-02 ds3000-03]" notUpdated=[]
15:18:31 INF Switches status notReady="[sse-c4632-01 ds3000-01 ds3000-02 ds3000-03]" notUpdated=[]
15:19:01 INF Switches status notReady="[sse-c4632-01 ds3000-01 ds3000-03]" notUpdated=[]
15:19:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:20:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:20:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:21:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:21:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:22:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:22:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:23:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:23:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:24:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:24:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:25:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:25:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:26:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:26:32 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:27:02 INF Switches status notReady=[ds3000-03] notUpdated=[]
15:27:31 ERR Initial test suite setup failed suite="Basic VPC Peering Suite" error="setting up VPCs: waiting for ready: waiting for ready: context deadline exceeded"
```

pau-hedgehog:
https://github.com/githedgehog/fabricator/actions/runs/22705183717/job/65831088128
```
[...]

PR #1553, opened 2026-03-11
fix(test): wait for switches after VPC deletion before recreating

pau-hedgehog:
Fixes https://github.com/githedgehog/fabricator/issues/1254

### DIFF
--- a/pkg/hhfab/testing.go
+++ b/pkg/hhfab/testing.go
@@ -901,6 +901,7 @@ func (c *Config) SetupVPCs(ctx context.Context, vlab *VLAB, opts SetupVPCsOpts) 
 	if err := kube.List(ctx, existingVPCs); err != nil {
 		return fmt.Errorf("listing existing VPCs: %w", err)
 	}
+	deletedVPCs := false
 	for _, vpc := range existingVPCs.Items {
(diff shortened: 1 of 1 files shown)


pkg/hhfab/testing.go
```diff
@@ -901,19 +901,30 @@ func (c *Config) SetupVPCs(ctx context.Context, vlab *VLAB, opts SetupVPCsOpts)
...
 			slog.Info("Deleted", "vpc", vpc.Name)
 			changed = true
+			deletedVPCs = true
 		}
 	}
 
 	if changed {
-		slog.Debug("Waiting for deletions to settle")
-		time.Sleep(3 * time.Second)
+		// Wait for switches to process deletions before creating new VPCs
```
edipascale: I'd recommend waiting the 3 seconds regardless if there was a change, and *then* do the WaitReady too if VPCs were deleted, to make sure that agents pick up the changes and we don't immediately return

pau-hedgehog:
@Frostman, could you please also review? We are hitting hit also on env-1 during routine RC testing:
```
admin@s5232-02:~$ grep -c "VNI is already used in VRF VrfVvpc-0" /var/log/agent.log 
49
```

Frostman (approved):

PR #1556, opened 2026-03-13
chore(vlab): add configurable server VM mtu and use 9036 for HLAB

pau-hedgehog:
## Before this PR

There is a `--interface-mtu` flag on `setup-vpcs` that lets users manually set an MTU value. Its default is 0, meaning "don't set any MTU."

Before this PR, when `setup-vpcs` ran with MTU=0 (the default), two things could happen depending on whether DNS/NTP servers were configured:

Without DNS/NTP: no DHCP options object created. DHCP hands out IP addresses but advertises no MTU. Servers use their OS default (1500).

With DNS/NTP (e.g. `--dns-servers 8.8.8.8`): a DHCP options object is created to carry the DNS servers. The VPC API sees MTU=0 in that object and auto-fills it to 9036. DHCP now advertises 9036 jumbo MTU to all servers, including SONiC VS where jumbo frames crash the VS. There was no protection against this.

Server interfaces themselves were never configured with any MTU either. They just used the OS default (1500). So even if DHCP advertised 9036, the server interfaces weren't set up for it.

Bottom line: MTU was not actively managed. It was either not advertised, or silently wrong (9036 advertised via DHCP on VS environments that can't handle it).

### Auto-configure MTU based on environment

When `--interface-mtu` is not explicitly set (defaults to 0), the code now detects what kind of switches are running and picks the right value:

Non-VS (HLAB/production): sets MTU to 9036. Jumbo frames work here, this is the correct value.

SONiC VS with DNS/NTP configured: sets MTU to 1500. This prevents the API from auto-filling 9036 into the DHCP options object that must exist to carry the DNS/NTP servers.

SONiC VS without DNS/NTP: leaves MTU at 0. No DHCP options object is created, so the API never fills in 9036. No MTU advertised. Servers use default 1500.

### Configure server interfaces

The MTU value is now passed to `hhnet.sh` via `--mtu=<value>` when configuring bonds and VLANs on server VMs. Before, server interfaces were never told what MTU to use. Now they match what DHCP advertises.

### CLI validation

The `--interface-mtu` flag now validates that the value is between 96 and 9036 (matching the VPC API limits), and is available on `vlab up` in addition to `setup-vpcs`.

## Why we cannot always leave MTU at 0 for SONiC VS

The ideal behavior for SONiC VS is "do nothing with MTU" -- don't advertise jumbo frames, let servers use their default 1500. Leaving `InterfaceMTU` at 0 achieves this, but only when no DHCP options object is created. The problem is that we cannot always avoid creating one.

When the user configures DNS or NTP servers, the code must create a `VPCDHCPOptions` object to carry those values (`pkg/hhfab/testing.go:804-811`):

```go
var dhcpOpts *vpcapi.VPCDHCPOptions
if len(opts.DNSServers) > 0 || len(opts.TimeServers) > 0 || opts.InterfaceMTU > 0 {
	dhcpOpts = &vpcapi.VPCDHCPOptions{
		DNSServers:   opts.DNSServers,
		TimeServers:  opts.TimeServers,
		InterfaceMTU: opts.InterfaceMTU,
	}
}
```

Once that object exists, the VPC API defaulting webhook sees `InterfaceMTU == 0` and sets it to 9036 (`vendor/go.githedgehog.com/fabric/api/vpc/v1beta1/vpc_types.go:306-317`):

```go
if subnet.DHCP.Options != nil {
	if subnet.DHCP.Options.InterfaceMTU == 0 {
		subnet.DHCP.Options.InterfaceMTU = 9036
	}
}
```

There is no way to create a DHCP options object with DNS servers without also triggering the MTU default. So the code explicitly sets 1500 to fill the slot before the API can inject 9036 (`pkg/hhfab/testing.go:607-608`):

```go
if len(opts.DNSServers) > 0 || len(opts.TimeServers) > 0 {
	opts.InterfaceMTU = 1500
}
```

When no DNS/NTP are configured, all three conditions in the `if` on line 805 are false, `dhcpOpts` stays `nil`, the API webhook never fires, and no MTU is advertised. Leaving `InterfaceMTU` at 0 is safe in this case.

Part of https://github.com/githedgehog/internal/issues/338

### DIFF
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -1135,6 +1135,12 @@ func Run(ctx context.Context) error {
 								Name:  FlagNameVPCMode,
 								Usage: "VPC mode to be used for on-ready commands: empty is default (l2vni), l3vni, etc.",
 							},
+							&cli.UintFlag{
+								Name:    "interface-mtu",
+								Aliases: []string{"mtu"},
+								Usage:   "interface MTU for server interfaces and VPCs advertised by DHCP when using on-ready commands (0 = auto: 9036 for non-VS, disabled for SONiC VS)",
+								EnvVars: []string{"HHFAB_INTERFACE_MTU"},
+							},
 							&cli.StringSliceFlag{
 								Name:    FlagReleaseTestRegexes,
 								Aliases: []string{"rt-regex"},
@@ -1212,6 +1218,11 @@ func Run(ctx context.Context) error {
 								onReady = append(onReady, ready)
 							}
 
+							ifMTU, err := parseInterfaceMTU(c)
+							if err != nil {
+								return err
+							}
+
 							if err := hhfab.VLABUp(ctx, workDir, cacheDir, hhfab.VLABUpOpts{
 								HydrateMode:          hhfab.HydrateMode(hydrateMode),
 								ReCreate:             c.Bool(FlagNameReCreate),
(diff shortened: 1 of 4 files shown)


edipascale (approved):
thanks for looking into my concerns and for the detailed explanation in the PR description!

Frostman (approved):

PR #1587, opened 2026-03-22
chore(diagram): add underlay layer for drawio

pau-hedgehog:
image

### DIFF
--- a/pkg/hhfab/cmddiagram.go
+++ b/pkg/hhfab/cmddiagram.go
@@ -11,6 +11,7 @@ import (
 
 	agentapi "go.githedgehog.com/fabric/api/agent/v1beta1"
 	dhcpapi "go.githedgehog.com/fabric/api/dhcp/v1beta1"
+	gwapi "go.githedgehog.com/fabric/api/gateway/v1alpha1"
 	vpcapi "go.githedgehog.com/fabric/api/vpc/v1beta1"
 	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
 	"go.githedgehog.com/fabric/pkg/util/kubeutil"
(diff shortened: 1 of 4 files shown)


Frostman (changes requested):
While it could be pretty useful in debug cases, I think it adds a lot of details that are mostly never needed other than for debugging. I suggest making it optional (and disabled by default) and having a flag for diagram generation to include underlay data. It would be useful to add ASNs too in that case.

pau-hedgehog:
> While it could be pretty useful in debug cases, I think it adds a lot of details that are mostly never needed other than for debugging. I suggest making it optional (and disabled by default) and having a flag for diagram generation to include underlay data. It would be useful to add ASNs too in that case.

My bad for not clarifying in the description. Underlay is a separate drawio layer which is not exposed by default. You can check yourself in the attached diagram generated with this branch
[diagram.drawio](https://github.com/user-attachments/files/27497816/diagram.drawio)

pau-hedgehog:
> > While it could be pretty useful in debug cases, I think it adds a lot of details that are mostly never needed other than for debugging. I suggest making it optional (and disabled by default) and having a flag for diagram generation to include underlay data. It would be useful to add ASNs too in that case.
> 
> My bad for not clarifying in the description. Underlay is a separate drawio layer which is not exposed by default. You can check yourself in the attached diagram generated with this branch [diagram.drawio](https://github.com/user-attachments/files/27497816/diagram.drawio)

@edipascale could you please review from here? Thanks

edipascale (approved):
I don't know much about drawio APIs but the fabric related code looks good and I've tested it on a live deployment, so LGTM

Frostman:
@pau-hedgehog, my bad, I keep forgetting about layers

Frostman (approved):

PR #1588, opened 2026-03-22
ci: add release upgrade choices for custom vlab workflow dispatch

### LINKED ISSUE #1585: Upgrade from version workflow dispatch free form field can contain invalid options
Image

https://github.com/githedgehog/fabricator/actions/runs/23362731817/job/67969774924
```
Inputs:
  Fabricator reference: refs/heads/master
  Pre-build script: echo 'Noop pre-build'
  Fabric mode: spine-leaf
  Mesh: true
  Gateway: true
  Include ONIE: false
  Build mode: iso
  VPC mode: l2vni
  Hybrid: true
  Upgrade from: Static External
  Release test: true
  Release test regex: 
  Release test regex invert: false
  Debug: true
  Pause on failure: true
Calculated:
  Slug: h-upStatic External-mesh-gw-iso-l2vni-rt
```

pau-hedgehog:
Fixes https://github.com/githedgehog/fabricator/issues/1585

image

### DIFF
--- a/.github/workflows/custom-vlab.yaml
+++ b/.github/workflows/custom-vlab.yaml
@@ -39,9 +39,14 @@ on:
         default: false
       upgradefrom:
         description: "Upgrade from version"
-        type: string
+        type: choice
         required: false
         default: ""
+        options:
+          - ""
+          - "25.03"
+          - "25.04"
+          - "25.05"
       releasetest:
         description: "Run Release tests"
         type: boolean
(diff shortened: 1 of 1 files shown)


Frostman (changes requested):
I think it's easy enough to type "25.05" in that upgradefrom field and it doesn't require maintenance - like updating the list after each release. More than that, we don't support upgrading from 25.03/25.04

PR #1632, opened 2026-04-05
Skip show tech when no failures

pau-hedgehog:


Frostman (changes requested):
You need to introduce another env var/option that would be to force collecting show, as right now, HHFAB_VLAB_COLLECT is used just to enable it, and I'd like to keep it. You can add HHFAB_VLAB_COLLECT_FORCE  if you want or smth like this. The rest of the logic makes sense to me

PR #1648, opened 2026-04-10
Add Host-BGP Release Test Suite

pau-hedgehog:


edipascale (changes requested):
I see a few issues here:
1. the added tests use a hostBGP server attached to a single leaf, while in reality the feature was implemented to support multihoming on the TH5; but release tests cannot test this because test-connectivity does not support it yet. Sure, some coverage is better than none, but if we were to add tests I'd like to make sure we use the feature as our customers do
2. this PR includes a fix for the hostBGP command that was added a separate commit in the on-ready test refactor, and which would create conflicts. I'd really like that PR to get in soon and then we would not need it here, but at the very least I'd cherry-pick that fix and keep it as its own commit
3. I don't understand why we are creating a separate test context here, where everywhere else we just use the one, maybe changing the parameters before starting the suite if need be. What you are doing is not wrong per se, but uniformity in the way we do things makes reviewing and code maintenance easier.

My personal preference would be to get the on ready test refactor in, then revisit this with multihoming in mind, but I'm open to discussions

pau-hedgehog:
> 1. the added tests use a hostBGP server attached to a single leaf, while in reality the feature was implemented to support multihoming on the TH5; but release tests cannot test this because test-connectivity does not support it yet. Sure, some coverage is better than none, but if we were to add tests I'd like to make sure we use the feature as our customers do

Important point. I raised this as is as a first automation attempt but we hit again with the test-connectivity road-block. We need a wider discussion

> 2. this PR includes a fix for the hostBGP command that was added a separate commit in the on-ready test refactor, and which would create conflicts. I'd really like that PR to get in soon and then we would not need it here, but at the very least I'd cherry-pick that fix and keep it as its own commit

Based on this and the previous point I think I'll move to draft until we can discuss further.

> 3. I don't understand why we are creating a separate test context here, where everywhere else we just use the one, maybe changing the parameters before starting the suite if need be. What you are doing is not wrong per se, but uniformity in the way we do things makes reviewing and code maintenance easier.

Ack. Thanks for the feedback

pau-hedgehog:
Closing in favor of https://github.com/githedgehog/fabricator/pull/1692

PR #1664, opened 2026-04-13
chore(hack): add script to clean up cancelled dependabot CI runs

pau-hedgehog:
This is because I get annoyed too easily by N dependabot cancelled workflows per dependabot PR on https://github.com/githedgehog/fabricator/actions and sometimes I feel the urge them to delete them (was doing manually until now)

It also pollutes outputs from https://github.com/githedgehog/lab/pull/364


Frostman (changes requested):
I'd suggest you keep it locally or commit it to a tools repo or a lab repo. As for your scripts to collect jobs info, IMO they should correctly handle that case w/o removing the jobs.

PR #1702, opened 2026-04-27
VLAB: run iperf3 service in server VMs

pau-hedgehog:
Also idea from @Frostman, adds some additional time saving on test connectivity avoiding having to start iperf3 N times per test/workflow.

Main savings come from #1697. Stacked onto #1697.

`h-gw-iso-l2vni-rt` `Run and test hybrid VLAB` step:
- master: `3h 7m 39s`
- #1697 (bidir): `2h 40m 48s`
- this PR (bidir + iperf3 daemon): `2h 33m 24s` (-34m15s, ~18%)

### DIFF
--- a/pkg/hhfab/rt_nat_tests.go
+++ b/pkg/hhfab/rt_nat_tests.go
@@ -18,7 +18,6 @@ import (
 	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
 	"go.githedgehog.com/fabric/pkg/util/apiutil"
 	"go.githedgehog.com/fabricator/pkg/util/sshutil"
-	"golang.org/x/sync/errgroup"
 	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -275,7 +274,7 @@ func (testCtx *VPCPeeringTestCtx) testNATGatewayConnectivity(
 				if err != nil {
 					return err
 				}
-				if ies := checkIPerf(ctx, testCtx.tcOpts, serverA, serverB, sshConfigs[serverA], sshConfigs[serverB], destIP, reachability, false); len(ies) > 0 {
+				if ies := checkIPerf(ctx, testCtx.tcOpts, serverA, serverB, sshConfigs[serverA], destIP, reachability, false); len(ies) > 0 {
 					iperfErrors = append(iperfErrors, ies...)
 				}
 			}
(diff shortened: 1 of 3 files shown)


pkg/hhfab/vlab_server_butane.tmpl.yaml
```diff
@@ -53,6 +53,34 @@ systemd:
...
+        After=docker.service network-online.target
+        Wants=network-online.target
+        Requires=docker.service
+
+        [Service]
+        Type=simple
+        # Block until the docker daemon is responsive and the toolbox image is
+        # local. Without these gates the main ExecStart races docker bring-up
+        # and image pull on first boot, producing many on-failure restarts
+        # before the service stabilizes. The pull retries until success so a
```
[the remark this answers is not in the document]
pau-hedgehog: Good catch -- prefixed `ExecStop` with `-` to tolerate the container already being gone (which is the common case since `--rm` plus the cleanup `ExecStartPre` usually leaves nothing for `docker stop` to find). Updated comment in the unit explains why.

pkg/hhfab/testing.go
```diff
@@ -2334,35 +2349,59 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
...
 
-						// Acquire toolbox mutexes for both servers to avoid concurrent toolbox usage
+						// iperf3 runs in toolbox only on the source side (the destination
+						// is served by a long-lived systemd-managed iperf3 daemon, see
+						// vlab_server_butane.tmpl.yaml), so only the source's toolbox
+						// mutex needs to gate concurrent toolbox usage with other tests.
 						toolboxMutexes[serverA].Lock()
 						defer toolboxMutexes[serverA].Unlock()
 						if err := checkToolboxLock(ctx, serverA, clientA, 2*time.Minute); err != nil {
 							return fmt.Errorf("checkToolbockLock: %w", err)
```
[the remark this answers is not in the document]
pau-hedgehog: Fixed -- s/checkToolbockLock/checkToolboxLock/ in the wrap.

pkg/hhfab/testing.go
```diff
@@ -2460,6 +2499,15 @@ const (
 	ReachabilityReasonGatewayPeering ReachabilityReason = "gateway-peering"
 )
 
+// isBidirSafeReason reports whether two reachable directions sharing this
+// reason can be collapsed into a single iperf3 --bidir session. Only fully
+// symmetric, non-NAT peering reasons qualify: gateway peerings are excluded
+// because they may carry NAT, and Reachability does not currently surface
+// per-pair NAT state. See iperf3 --bidir notes in the PR description.
```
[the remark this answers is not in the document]
pau-hedgehog: Deferring to a follow-up. End-to-end proof for this PR is the HLAB CI run on the parent (PR #1697) plus local VLAB validation on this branch -- per-direction throughputs parsed cleanly, bidir gate fired on the right pairs, daemon mode passed conns. A fixture-based unit test would harden against iperf3 JSON-field drift but isn't strictly required for this change to land.

pkg/hhfab/rt_nat_tests.go
```diff
@@ -979,51 +978,13 @@ func (testCtx *VPCPeeringTestCtx) testPortForwardInboundConnectivity(
...
-					return fmt.Errorf("iperf3 client from %s to %s:%d: %w: %s", serverAName, natIP, externalPort, err, stderr)
-				}
-
-				return nil
-			})
-
-			if err := g.Wait(); err != nil {
-				return err
+			// iperf3 server is the long-lived systemd-managed daemon on serverB
+			// (see vlab_server_butane.tmpl.yaml), so the client just connects to
```
Frostman: Likely outdated comment referencing changes to `vlab_server_butane.tmpl.yaml`

pkg/hhfab/testing.go
```diff
@@ -3076,143 +3119,74 @@ func runIPerf3Test(ctx context.Context, opts TestConnectivityOpts, from, to stri
...
-		if opts.IPerfsTOS > 0 {
-			cmd += fmt.Sprintf(" --tos %d", opts.IPerfsTOS)
-		}
-		stdout, stderr, err := retrySSHCmd(ctx, fromSSH, cmd, from)
-		report, parseErr := parseIPerf3Report([]byte(stdout))
-		if err != nil {
-			if parseErr == nil && report.Error != "" {
-				fwd.ClientMsg = report.Error
+	// iperf3 server runs as a long-lived systemd-managed docker container on
+	// every server VM (see vlab_server_butane.tmpl.yaml), listening on the
```
Frostman: likely same here

Frostman (changes requested):

pkg/hhfab/rt_nat_tests.go
pau-hedgehog: Good catch -- comment was a leftover from an earlier draft that did install the unit via the butane template. Dropped it entirely; the surrounding code (no `iperf3 -s` startup, just the client connecting to the NAT IP) is self-evident. Pushed in 87a0689c.

pkg/hhfab/testing.go
pau-hedgehog: Same issue here -- trimmed the block to a single line that captures the only surprising bit (no server-side setup; the daemon listens on 5201 on every server VM). Pushed in 87a0689c.

pkg/hhfab/testing.go
```diff
@@ -2397,19 +2395,17 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
 						}
 						defer iperfs.Release(1)
 
-						// Acquire toolbox mutexes for both servers to avoid concurrent toolbox usage
+						// iperf3 runs in toolbox only on the source side (the destination
+						// is served by a long-lived systemd-managed iperf3 daemon, see
+						// vlab_server_butane.tmpl.yaml), so only the source's toolbox
```
edipascale, line 2376: nit: outdated comment about butane template

pkg/hhfab/testing.go
```diff
@@ -2984,14 +2980,61 @@ func checkToolboxLock(ctx context.Context, server string, ssh *sshutil.Config, t
 	return nil
 }
 
+// iperf3DaemonUnit is the systemd service definition for the always-on iperf3
```
edipascale: this is a neat trick to avoid issues in the upgrade, but shouldn't we also add it to the butane template, so that when eventually we move past an oldest base release without the template we can get rid of the function below? Or would it cause issues if we did?

edipascale (dismissed):
minor comments below but otherwise LGTM

pkg/hhfab/testing.go
Frostman: I think we should so we can remove the `ensure` thing after another release

Frostman (changes requested):
I think as Emanuele pointed, we should have it in a butane too while keeping the ensure function with the comment that it could be removed after we only support upgrades from 26.03

pkg/hhfab/testing.go
pau-hedgehog: addressed

pkg/hhfab/testing.go
pau-hedgehog: Adding iperf3.service to vlab_server_butane.tmpl.yaml for fresh installs; ensureIperf3Daemon kept as upgrade trick for VMs that pre-date the butane entry

edipascale (approved):

Frostman (approved):

PR #1714, opened 2026-05-03
chore(test): iperf3 below-floor diagnostics for gateway-peering investigation

pau-hedgehog:


Frostman:
Closed as stale, not needed atm

PR #1730, opened 2026-05-08
feat(hhfab): auto-derive VPC mode from switch profile

pau-hedgehog:
26.02 lifted the same-mode constraint on VPCPeering (fabric ddc72cc).
Requiring users to declare l2vni / l3vni is leftover friction now that
the only reason mode matters is the celestica-ds5000 L2VNI gap.

- SetupVPCs auto-derives each VPC's mode from SwitchProfile.Features.L2VNI
  of the attached leaves. `--vpc-mode` stays as a force-everywhere
  override (l3vni / l3flat).
- release-test reads per-VPC mode from VPC.Spec.Mode instead of the global
  setupOpts: ESLAG filter, per-VPC L3 waits, RequireAllServers relax.
- on-ready setup-peerings only references VPCs that exist after SetupVPCs.

Behavior change: `--vpc-mode=l2vni` (and the empty default) no longer
force L2VNI; they auto-derive. Identical on L2VNI-capable hardware;
mixed-mode VPCs on hardware with DS5000-only leaves.

Related: https://github.com/githedgehog/lab/pull/392

### DIFF
--- a/pkg/hhfab/rt_single_vpc_suite.go
+++ b/pkg/hhfab/rt_single_vpc_suite.go
@@ -162,21 +162,44 @@ func mclagTest(ctx context.Context, testCtx *VPCPeeringTestCtx) (bool, []RevertF
 // For each eslag connection, set one of the links down by shutting down the port on the switch,
 // then test connectivity. Repeat for the other link.
 func eslagTest(ctx context.Context, testCtx *VPCPeeringTestCtx) (bool, []RevertFunc, error) {
-	// l3vni mode is not compatible with ESLAG, so there will be no servers attached to ESLAG connections
-	if testCtx.setupOpts.VPCMode == vpcapi.VPCModeL3VNI {
-		return true, nil, fmt.Errorf("L3VNI mode is not compatible with ESLAG") //nolint:goerr113
-	}
 	// list connections in the fabric, filter by ES-LAG connection type
 	conns := &wiringapi.ConnectionList{}
 	if err := testCtx.kube.List(ctx, conns, kclient.MatchingLabels{wiringapi.LabelConnectionType: wiringapi.ConnectionTypeESLAG}); err != nil {
 		return false, nil, fmt.Errorf("listing connections: %w", err)
 	}
-	if len(conns.Items) == 0 {
-		slog.Info("No ESLAG connections found, skipping test")
+	// ESLAG is not compatible with L3VNI / L3Flat: SetupVPCs skips those
+	// servers from VPC creation. Keep only ESLAG conns whose attached server
+	// landed in an L2VNI VPC; skip the test if none remain.
+	attachList := &vpcapi.VPCAttachmentList{}
+	if err := testCtx.kube.List(ctx, attachList); err != nil {
+		return false, nil, fmt.Errorf("listing VPCAttachments: %w", err)
+	}
+	vpcs := &vpcapi.VPCList{}
+	if err := testCtx.kube.List(ctx, vpcs); err != nil {
+		return false, nil, fmt.Errorf("listing VPCs: %w", err)
+	}
+	vpcModes := map[string]vpcapi.VPCMode{}
+	for _, vpc := range vpcs.Items {
+		vpcModes[vpc.Name] = vpc.Spec.Mode
+	}
+	l2vniConn := map[string]bool{}
+	for _, attach := range attachList.Items {
+		if mode, ok := vpcModes[attach.Spec.VPCName()]; ok && mode == vpcapi.VPCModeL2VNI {
+			l2vniConn[attach.Spec.Connection] = true
+		}
+	}
+	eligible := make([]wiringapi.Connection, 0, len(conns.Items))
+	for _, conn := range conns.Items {
+		if l2vniConn[conn.Name] {
+			eligible = append(eligible, conn)
+		}
+	}
+	if len(eligible) == 0 {
+		slog.Info("No ESLAG connections with L2VNI-attached servers found, skipping test")
 
 		return true, nil, errNoEslags
 	}
-	for _, conn := range conns.Items {
+	for _, conn := range eligible {
 		slog.Debug("Testing ESLAG connection", "connection", conn.Name)
 		if len(conn.Spec.ESLAG.Links) != 2 {
 			return false, nil, fmt.Errorf("ESLAG connection %s has %d links, expected 2", conn.Name, len(conn.Spec.ESLAG.Links)) //nolint:goerr113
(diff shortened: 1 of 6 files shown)


pau-hedgehog:
Tested also on env-4 and env-5:
```
pau@pau-TP-P14s:/tmp/env-ctl-import/env-ctl$ ./env-ctl.sh status
=== matrix 20260525-081747-matrix-6a039ea26 ===
hhfab:    pau/test-l2vni-l3vni-peering @ 6a039ea26
wiring:   pau/env-5-mixed-vpc @ e9a7257
iperfs-speed: env-4=7000

env         topology       mode     phase           result                                runtime  last-event (CEST)
env-4       spine-leaf     l2vni    done            release - PASS 21/45 (24 skip)        2h24m    12:43 CEST run end rc=0
env-5       spine-leaf     l3vni    done            release - FAIL 38/45 (1 fail, 6 skip) (exit 1)  1h37m    11:57 CEST run end rc=1
env-5       mesh           l3vni    done            release - PASS 35/45 (10 skip)        1h15m    13:13 CEST run end rc=0
env-5       mesh           l3vni    done            iperf - PASS 4/4                      6m2s     13:19 CEST run end rc=0

failures in env-5 / spine-leaf / l3vni (from attempt 1, 11:57 CEST):
- RoCE flag and basic traffic marking: no UC3 transmit bits/pkts found on switch ds3000-06 after 6 attempts
  show-tech: ssh env-5 ls /home/hedgehog/hhfab/show-tech-output/single-vpc-suite/roce-flag-and-basic-traffic-marking/
drill: ssh env-5 less env-ctl-runs/20260525-081747-matrix-6a039ea26-spine-leaf-release/release-test.log
```
The env-5 spine-leaf run hit: https://github.com/githedgehog/fabricator/issues/1755

env-5/spine-leaf is the only topology where we have mixed l2VNI/l3VNI VPCs. env-4 is all L2VNI and env-5/mesh is all L3VNI:
image
image
image

pau-hedgehog:
@coderabbitai review

pkg/hhfab/rt_single_vpc_suite.go
```diff
@@ -128,21 +128,44 @@ func makeSingleVPCSuite() *JUnitTestSuite {
...
-	// l3vni mode is not compatible with ESLAG, so there will be no servers attached to ESLAG connections
-	if testCtx.setupOpts.VPCMode == vpcapi.VPCModeL3VNI {
-		return true, nil, fmt.Errorf("L3VNI mode is not compatible with ESLAG") //nolint:goerr113
-	}
 	// list connections in the fabric, filter by ES-LAG connection type
 	conns := &wiringapi.ConnectionList{}
 	if err := testCtx.kube.List(ctx, conns, kclient.MatchingLabels{wiringapi.LabelConnectionType: wiringapi.ConnectionTypeESLAG}); err != nil {
 		return false, nil, fmt.Errorf("listing connections: %w", err)
 	}
-	if len(conns.Items) == 0 {
```
edipascale, line 140: I'd keep this - if there are 0 ESLAG cnonections there's no need to do anything else regardless, right?

pkg/hhfab/vlabrunner.go
```diff
@@ -702,10 +702,43 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 						return c.handleShutdownWithPause(ctx, vlab, opts, fmt.Errorf("setting up VPCs: %w", err))
 					}
 				case OnReadySetupPeerings:
-					// TODO make it configurable
-					peerings := []string{"1+2"}
-					if c.Fab.Spec.Config.Gateway.Enable {
-						peerings = append(peerings, "2+3:gw:vpc1=subnet-01:vpc2=subnet-01")
+					// Build peering requests defensively: only reference VPCs that
```
edipascale: I'm not sure about this; it results in different things being tested compared to what we expected / requested, which is not reassuring. If something in CI currently fails to meet the bar of entry, we can probably adapt the vlab generation to increase the number of servers etc. to make sure we have enough; if it happens on a custom setup then I think it's fine getting a hard failure

pkg/hhfab/testing.go
```diff
@@ -713,9 +791,61 @@ func (c *Config) SetupVPCs(ctx context.Context, vlab *VLAB, opts SetupVPCsOpts)
...
+			}
+			mode = derived
+		}
+		serverModes[server.Name] = mode
+		if !seenMode[mode] {
+			seenMode[mode] = true
+			modeOrder = append(modeOrder, mode)
+		}
+	}
+	// L2VNI is both the zero value and the only "unforced" VPCMode, so an
```
edipascale, line 835: I'm a little conflicted about the new meaning of "l2vni". I understand that in practice there's not much sense in keeping the old behavior, at least at the moment, but I wish "l2vni" actually meant l2vni and the empty default meant "auto" or something, which would be the new behavior. This is probably not worth spending time or tokens on, but I hadto say it 😛

edipascale (changes requested):
One nit, one philosophical rant and one actual request for change (the flexible peerings)

pkg/hhfab/rt_single_vpc_suite.go
pau-hedgehog: Restored in 5bf0cb2d0: the test returns early when there are no ESLAG connections at all, before listing VPCs and attachments.

pkg/hhfab/vlabrunner.go
pau-hedgehog: Agreed, reverted in 0ae599058, on-ready setup-peerings requests the canonical 1+2 / 2+3:gw again.

To make that an actual hard failure rather than a dangling request, SetupPeerings now rejects any request naming a VPC that does not exist: the gateway path used to fall through to an empty expose instead of erroring.

CI already avoids the case that motivated the defensive code, run-vlab.yaml sets VLAB_ESLAG_GROUPS="" for l3vni runs so no server is skipped and all three VPCs exist.

pkg/hhfab/testing.go
pau-hedgehog: Turned out cheap, so it is in b675e83d1: `--vpc-mode=auto` (and unset) derives per VPC, `--vpc-mode=l2vni` forces L2VNI everywhere.

`vpcapi.VPCModeL2VNI` is `""` upstream, so the opts structs carry a separate `AutoVPCMode` bool and the CLI maps the flag onto both.

edipascale: sorry, but this looks like a hack to me - you cannot define a new "auto" vpcapi mode here, because they belong in fabric, and the result is that you have to carry two flags, which is even more confusing and can create more issues than it solves. I appreciate you trying to placate my OCD but I don't think this is the right way, that's why I tagged it as a rant - changing this is beyond the scope of this PR and probably not worth it

edipascale (approved):

Frostman (approved):

PR #1747, opened 2026-05-13
hostBGP release-test

### LINKED ISSUE #1359: Basic host bgp release test

edipascale:
Fix #1359

### DIFF
--- a/pkg/hhfab/matrix.go
+++ b/pkg/hhfab/matrix.go
@@ -633,11 +633,22 @@ func runMatrixServerServerPhase(ctx context.Context, opts TestConnectivityOpts, 
 				}
 			}
 
+			// a /32 VIP on lo is never the kernel's own choice of source, and a
+			// host in several hostBGP VPCs has one VIP per VPC. Only this phase
+			// binds it: the proto-port, port-forward and curl phases key probes
+			// by server name, which cannot name one VIP out of several, so a
+			// hostBGP source there probes from whatever the kernel picks.
+			srcIP := netip.Addr{}
+			if src.Server.HostBGP {
+				srcIP = src.Server.IP
+			}
+
 			args := pingIperfPairArgs{
 				From:     src.Server.Name,
 				To:       dst.Server.Name,
 				FromSSH:  deps.sshByServer[src.Server.Name],
 				ToIP:     toIP,
+				SrcIP:    srcIP,
 				Expected: expected,
 				Bidir:    bidir,
 				Pings:    deps.pings,
--- a/pkg/hhfab/rt_no_vpc_suite.go
+++ b/pkg/hhfab/rt_no_vpc_suite.go
@@ -9,7 +9,9 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"maps"
 	"net/http"
+	"net/netip"
 	"net/url"
 	"slices"
 	"strconv"
@@ -18,9 +20,13 @@ import (
 
 	agentapi "go.githedgehog.com/fabric/api/agent/v1beta1"
 	gwapi "go.githedgehog.com/fabric/api/gateway/v1alpha1"
+	vpcapi "go.githedgehog.com/fabric/api/vpc/v1beta1"
 	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
+	"go.githedgehog.com/fabric/pkg/hhfctl"
 	fabricatorapi "go.githedgehog.com/fabricator/api/fabricator/v1beta1"
+	"go.githedgehog.com/fabricator/pkg/fab"
 	"go.githedgehog.com/fabricator/pkg/fab/comp"
+	"go.githedgehog.com/fabricator/pkg/util/sshutil"
 	"golang.org/x/sync/errgroup"
 	corev1 "k8s.io/api/core/v1"
 	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -65,6 +71,14 @@ func makeNoVpcsSuite() *JUnitTestSuite {
 			F:    prometheusObservabilityTest,
 			SkipFlags: SkipFlags{
 				NoProm: true,
+			},
+		},
+		{
+			Name: "HostBGP Multihoming",
+			F:    hostBGPTest,
+			SkipFlags: SkipFlags{
+				SubInterfaces: true,
+				NoServers:     true,
 			},
 		},
 	}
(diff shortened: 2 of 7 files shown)


pkg/hhfab/rt_utils.go
```diff
@@ -1250,3 +1252,132 @@ func findAnyAttachedServer(ctx context.Context, kube kclient.Client) (*AttachedS
...
+			continue
+		}
+		switches := map[string]bool{}
+		for _, conn := range connList.Items {
+			if conn.Spec.Unbundled != nil {
+				switches[conn.Spec.Unbundled.Link.Switch.DeviceName()] = true
+			}
+		}
+		if len(switches) >= 2 {
+			mhs := MultiHomedServer{
```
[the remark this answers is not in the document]
edipascale: irrelevant, we are using unbundled connections, not mclag ones

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,503 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+			allAttaches = append(allAttaches, makeVPCAttachment(regularVPCA.Connection.Name, vpcName, regularSubnetID))
+		case regularVPCBName:
+			regularVPCB.VPC = regularVPC
+			allAttaches = append(allAttaches, makeVPCAttachment(regularVPCB.Connection.Name, vpcName, regularSubnetID))
+		}
+	}
+
+	// Register cleanup revert before creating any resources
+	reverts = append(reverts, func(ctx context.Context) error {
+		return hhfctl.VPCWipeWithClient(ctx, testCtx.kube)
```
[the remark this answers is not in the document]
edipascale: not particularly relevant, as that config does nothing without matching config on the leaves, and any new VPC creation will reset the host config

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,503 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+				return fmt.Errorf("parsing VIP %q: %w", line, err)
+			}
+			switch {
+			case hostBGP.A.Subnet.Contains(prefix.Addr()):
+				hostBGP.A.VIP = prefix
+			case hostBGP.B.Subnet.Contains(prefix.Addr()):
+				hostBGP.B.VIP = prefix
+			default:
+				slog.Warn("hostBGP VIP does not match any of the VPCs subnets, ignoring it", "VIP", prefix.String())
+			}
```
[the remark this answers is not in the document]
edipascale: the VIP is configured manually by FRR on the host and it is nearly instantaneous, the retry is overkill even in the main branch

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,503 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
 
 	return nil
 }
+
+// prerequisites: no existing VPCs, at least 1 unbundled multihomed server, at least 2 other servers.
+// look for a server with unbundled connections to different switches; create two separate hostBGP VPCs,
+// and attach them both twice to the server via the two connections.
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,503 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+			continue
+		}
+
+		if regularVPCA.Server == nil {
+			regularVPCA.Server = &server
+			regularVPCA.Connection = &connList.Items[0]
+		} else {
+			regularVPCB.Server = &server
+			regularVPCB.Connection = &connList.Items[0]
+
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/testing.go
```diff
@@ -430,42 +430,63 @@ func GetServerNetconfCmd(conn *wiringapi.Connection, opts ServerNetconfOpts) (st
...
 }
 
-// TODO: multi subnet support once test-connectivity supports it
-func getServerHostBGPCmd(conn *wiringapi.Connection, vlan uint16, subnet netip.Prefix, serversInSubnet int) (string, error) {
-	if conn == nil {
-		return "", fmt.Errorf("connection is nil")
+type HostBGPParams struct {
+	VPCLabel     string
+	Connections  []*wiringapi.Connection
+	Vlan         uint16
```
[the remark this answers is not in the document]
edipascale: fixed

Frostman:
@coderabbitai review

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,521 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+			}
+			switch {
+			case hostBGP.A.Subnet.Contains(prefix.Addr()):
+				hostBGP.A.VIP = prefix
+			case hostBGP.B.Subnet.Contains(prefix.Addr()):
+				hostBGP.B.VIP = prefix
+			default:
+				slog.Warn("hostBGP VIP does not match any of the VPCs subnets, ignoring it", "VIP", prefix.String())
+			}
+		}
```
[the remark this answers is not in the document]
edipascale: not a real issue, addresses are assigned locally on the loopback via FRR config and there is no BGP convergence

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,521 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+		return changeAgentStatus(ctx, failoverSwSSH, failoverSwitchName, true)
+	})
+
+	slog.Debug("Shutting down link for failover test", "switch", failoverSwitchName, "port", failoverNOSPort)
+	if err := changeSwitchPortStatus(ctx, failoverSwSSH, failoverSwitchName, failoverNOSPort, false); err != nil {
+		return false, reverts, fmt.Errorf("shutting down port %s on %s: %w", failoverNOSPort, failoverSwitchName, err)
+	}
+	reverts = append(reverts, func(ctx context.Context) error {
+		return changeSwitchPortStatus(ctx, failoverSwSSH, failoverSwitchName, failoverNOSPort, true)
+	})
```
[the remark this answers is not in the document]
edipascale: addressed. no need to reply

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -981,3 +994,521 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		return false, reverts, fmt.Errorf("error configuring networking on servers: %w", err)
+	}
+
+	slog.Debug("Waiting for network convergence...")
+	time.Sleep(30 * time.Second)
```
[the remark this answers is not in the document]
edipascale: addressed. no need to reply

pkg/hhfab/rt_utils.go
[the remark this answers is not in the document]
pau-hedgehog: The on-ready suite guards against servers with unbundled connections to MCLAG-peered switches via `!isMclag`. Was that dropped intentionally here?

pau-hedgehog (dismissed):
Very minor comment/question but LGTM

edipascale:
@pau-hedgehog @Frostman I had added an extra commit to recycle an ESLAG connection as a series of unbundled connections for the purpose of this test, to allow it to run in the standard topologies we generate. Unfortunately reverting this change bricks the eslag leaves due to a gnmi ordering issue:
```
time=2026-06-19T12:22:05.786Z level=DEBUG msg=Action idx=74 weight=25 summary="Update Interface Ethernet1 Ethernet Base" command=update path="/interfaces/interface[name=Ethernet1]/ethernet"
time=2026-06-19T12:22:05.917Z level=ERROR msg=Failed err="failed to run agent: failed to process agent config from k8s: failed to process agent config loaded from k8s: failed to apply actions: gNMI set request failed: gnmi set request failed: rpc error: code = InvalidArgument desc = Cannot configure Ethernet1 as member of PortChannel2 as sub-interface is created on Ethernet1."
```

I've removed the commit; we can always add it back later on, either by fixing the gnmi ordering (ideally) or by simply not reverting the state (not an issue given where the test is placed in the rt suites right now, but I like to always have tests leave the system as they found it in case we ever move them around in the future)

pau-hedgehog:
> @pau-hedgehog @Frostman I had added an extra commit to recycle an ESLAG connection as a series of unbundled connections for the purpose of this test, to allow it to run in the standard topologies we generate.

So how do we handle this? Do we create specific wirings to test this such as https://github.com/githedgehog/lab/pull/420 ? Otherwise this can't be tested as often as it should

edipascale:
> So how do we handle this? Do we create specific wirings to test this such as [githedgehog/lab#420](https://github.com/githedgehog/lab/pull/420) ? Otherwise this can't be tested as often as it should

my idea is to have the standard CI vlab job also generate a multihomed server now that they are supported by SetupVPCs, but I did not want to introduce a potentially breaking change like this right before the release

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+		}
+		reverts = append(reverts, func(ctx context.Context) error {
+			return changeSwitchPortStatus(ctx, failoverSwSSH, failoverSwitchName, failoverNOSPort, true)
+		})
+	}
+
+	slog.Debug("Waiting for BGP reconvergence after link failover...")
+	time.Sleep(30 * time.Second)
+
+	if err := DoVLABTestConnectivityWithMatrix(ctx, testCtx.vlabCfg.WorkDir, testCtx.vlabCfg.CacheDir, testCtx.tcOpts, matrix); err != nil {
```
pau-hedgehog: The failover half of the test cannot fail.

After shutting the ports the test re-runs the identical matrix, and nothing asserts that the ports went down. In a passing run the only two mentions of the port are the two commands themselves:

```
14:53:41 DBG Changing switch port status device=leaf-02 port=Ethernet4 up=false
14:54:21 DBG Changing switch port status device=leaf-02 port=Ethernet4 up=true
```

`changeSwitchPortStatus` issues `shutdown` through the sonic-cli and returns success as soon as the command is accepted. The two connectivity results either side of it are indistinguishable, `took=9.036874724s` before and `took=8.68589812s` after, so if the shutdown had no effect the test would still pass and report exactly what it reports now.

Reading the operational status of each shut port before the second connectivity check would close this. Asserting that the next hop set for the loopback address on the peer leaf shrank would be stronger.

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+		if err != nil {
+			return fmt.Errorf("error getting hostBGP command: %w", err)
+		}
+		slog.Debug("Starting hostBGP container", "server", mhServer.Server.Name, "args", cmd)
+		fullCmd := fmt.Sprintf("docker run --network=host --privileged --rm --detach --name hostbgp ghcr.io/githedgehog/host-bgp:%s %s", fab.Versions.Platform.HostBGPContainer, cmd)
+		_, stderr, err := ssh.Run(ctx, fullCmd)
+		if err != nil {
+			return fmt.Errorf("starting hostbgp on %s: %w: %s", mhServer.Server.Name, err, stderr)
+		}
+		time.Sleep(1 * time.Second)
```
pau-hedgehog, line 1345: One second and a single read, for two addresses.

```go
		time.Sleep(1 * time.Second)
		stdout, stderr, err := ssh.Run(ctx, "/opt/bin/hhnet getvips")
```

One second was enough in the runs I did, so this is not an observed flake. The concern is that both existing hostBGP paths poll and this one does not: `rt_on_ready_suite.go:562` retries ten times at two second intervals, `testing.go:1108` five times at one second intervals. This test waits for two loopback addresses rather than one, so it has less slack than either, and a miss fails the test outright at the `IsValid()` check below.

The missing retry is also cheaper to add than to diagnose later, because there is no log line recording what `getvips` returned. The test logs `Starting hostBGP container` and then nothing until convergence, so a failure here gives you the error text and no evidence of what the server actually had.

Related, in the same loop: any line of `hhnet getvips` output that is not a prefix kills the test through `return fmt.Errorf("parsing VIP %q: %w", line, err)`. The on-ready loop ignores unparseable lines instead.

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+			})
+		}
+	}
+
+	if err := DoVLABTestConnectivityWithMatrix(ctx, testCtx.vlabCfg.WorkDir, testCtx.vlabCfg.CacheDir, testCtx.tcOpts, matrix); err != nil {
+		return false, reverts, fmt.Errorf("connectivity check failed: %w", err)
+	}
+
+	// Link failover: bring down all connections to one of the switches
+	var failoverSwitchName string
```
pau-hedgehog: The switch to fail over is picked at random.

```go
	var failoverSwitchName string
	for k := range mhServer.SwitchConns {
		failoverSwitchName = k

		break
	}
```

Go randomises map iteration order, so consecutive runs on the same topology can shut different switches and a failure will not reproduce. Sort the keys, take the first, and log which one was chosen.

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+		}
+		// deleting and recreating VPCs too quickly upsets the switches (fabric#656)
+		time.Sleep(3 * time.Second)
+		if err := WaitReady(ctx, testCtx.kube, testCtx.wrOpts); err != nil {
+			return false, reverts, fmt.Errorf("waiting for switches after wiping VPCs: %w", err)
+		}
+	}
+
+	// Register cleanup revert before creating any resources
+	reverts = append(reverts, func(ctx context.Context) error {
```
pau-hedgehog, line 1242: The servers are left configured.

The reverts cover the VPC wipe, re-enabling the agent and bringing the ports back up. The privileged `hostbgp` container keeps running on the multihomed server and the two regular servers keep their VLAN interfaces. In a full suite run the next `SetupVPCs` cleans this up at `testing.go:1097`, but a run scoped to this test alone leaves a container advertising host routes into the fabric. A revert running `docker stop -t 1 hostbgp` and `hhnet cleanup` on the three servers is cheap.

pkg/hhfab/matrix.go
```diff
@@ -517,11 +517,19 @@ func runMatrixServerServerPhase(ctx context.Context, opts TestConnectivityOpts,
 				}
 			}
 
+			// a /32 VIP on lo is never the kernel's own choice of source, and a
```
pau-hedgehog, line 636: Source address binding reaches one phase only.

```go
			srcIP := netip.Addr{}
			if src.Server.HostBGP {
				srcIP = src.Server.IP
			}
```

This applies to every matrix run, but only `runMatrixServerServerPhase` passes `SrcIP` on. `runMatrixProtoPortPhase` still passes `nil` for the source address at `matrix.go:762`, and `checkTCPPort`, `checkUDPPort`, the curl phase and `runMatrixIperfPortForward` all let the kernel choose. Nothing is broken today because no suite has a hostBGP source with protocol and port entries, so a comment recording the gap is enough if threading it through is out of scope.

pkg/hhfab/rt_single_vpc_suite.go
```diff
@@ -998,33 +980,27 @@ func dnsNtpMtuTest(ctx context.Context, testCtx *VPCPeeringTestCtx, matrix *Conn
...
 		DisableDefaultRoute: l3mode,
 	}
 
-	subnet.DHCP = vpcapi.VPCDHCP{
+	serverInfo.Subnet.DHCP = vpcapi.VPCDHCP{
 		Enable:  true,
 		Options: dhcpOpts,
 	}
-	change, err := CreateOrUpdateVpc(ctx, testCtx.kube, vpc)
+	change, err := CreateOrUpdateVpc(ctx, testCtx.kube, serverInfo.VPC)
```
pau-hedgehog: Wrapping an error that can be nil.

```go
	change, err := CreateOrUpdateVpc(ctx, testCtx.kube, serverInfo.VPC)
	if err != nil || !change {
		return false, nil, fmt.Errorf("updating VPC %s: %w", serverInfo.VPCName, err)
	}
```

When the desired DHCP options are already in place, `change` is false and `err` is nil, and the message renders as `updating VPC vpc-01: %!w(<nil>)`. Predates this pull request, but the line is being touched.

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+	}
+
+	g := errgroup.Group{}
+	g.Go(func() error {
+		ssh, ok := sshClients[mhServer.Server.Name]
+		if !ok {
+			return fmt.Errorf("no SSH client for %s", mhServer.Server.Name) //nolint:err113
+		}
+		// Cleanup any previous config
+		if _, _, err := ssh.Run(ctx, "docker stop -t 1 hostbgp"); err != nil {
```
pau-hedgehog:
```go
		if _, _, err := ssh.Run(ctx, "docker stop -t 1 hostbgp"); err != nil {
			// Ignore – container may not be running
			_ = err
		}
```

`_, _, _ = ssh.Run(ctx, "docker stop -t 1 hostbgp")` drops the dead branch and matches the existing idiom at `testing.go:1097`.

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+
+	for _, regularVPC := range []*RegularVPC{&regularVPCA, &regularVPCB} {
+		regularVPC.Endpoint = &Endpoint{Server: &ServerEndpoint{
+			Name:   regularVPC.Server.Name,
+			VPC:    regularVPC.VPC.Name,
+			Subnet: regularSubnetID,
+		}}
+	}
+
+	// the suite runs with noSetup, so VPCs left over by an earlier suite are still
```
pau-hedgehog, line 1223:
```go
	// the suite runs with noSetup, so VPCs left over by an earlier suite are still
	// around and their subnets would collide with the ones allocated above
```

The No VPCs suite runs first, at `rt_base.go:1041`, so there is no earlier suite. The wipe is guarding against a dirty environment left by a previous run, which is worth saying instead.

pkg/hhfab/rt_no_vpc_suite.go
```diff
@@ -998,3 +1011,513 @@ func (testCtx *VPCPeeringTestCtx) checkGatewayMetrics(ctx context.Context, prome
...
+		failoverNOSPort, ok := failoverPortMap[failoverLink.Switch.LocalPortName()]
+		if !ok {
+			return false, reverts, fmt.Errorf("port %s not in profile %s for switch %s", failoverLink.Switch.LocalPortName(), failoverProfile.Name, failoverSwitchName) //nolint:goerr113
+		}
+		if err := changeSwitchPortStatus(ctx, failoverSwSSH, failoverSwitchName, failoverNOSPort, false); err != nil {
+			return false, reverts, fmt.Errorf("shutting down port %s on %s: %w", failoverNOSPort, failoverSwitchName, err)
+		}
+		reverts = append(reverts, func(ctx context.Context) error {
+			return changeSwitchPortStatus(ctx, failoverSwSSH, failoverSwitchName, failoverNOSPort, true)
+		})
```
pau-hedgehog, line 1501: That one does not apply. `failoverNOSPort` is declared with `:=` inside the loop body, so Go scopes it per iteration and each revert closure gets its own port. No change needed.

pau-hedgehog (commented):
The refactor half looks good. Extracting `vpcSubnetAllocator`, `makeVPCAttachment` and `findUnbundledMHServers` into `rt_utils.go` is a clean deduplication and the on-ready rewrite is mechanical. The third commit fixes real bugs: the DHCP config is backed up with `DeepCopy` and restored on the one subnet instead of blanket-resetting every subnet in the VPC, hostBGP subnets are skipped when looking for an attached server, and the matrix endpoint is rebound after the hhnet cycle so a new DHCP address does not poison later pairs.

The hand-built matrix is justified rather than a shortcut. `populateConnectivityMatrix` calls `IsServerReachable(ctx, kube, src.Server.Name, dst.Server.Name, gatewayEnabled)` at `pkg/hhfab/testing.go:1916`, keyed on server name alone, so the generated oracle would mark `hostbgp-02` to `regular-01` reachable and erase the isolation half of this test. Worth filing that oracle limitation as its own issue so the workaround has something to point at.

I checked the findings below against a run where the test actually executes and passes, so none of them are speculative failures. The source address binding works as intended, visible in that run as `Running ping from=server-09 to=10.0.2.2 sourceIP=10.0.0.1 expected=true`. See the separate comment for what it takes to get the test running in CI.

Comments below, roughly in order of how much they matter.

pau-hedgehog:
The test skips in every VLAB job on this pull request:

```
07:39:05 INF * Running test test="HostBGP Multihoming"
07:39:05 WRN SKIP test="HostBGP Multihoming" reason="no available unbundled multihomed server"
```

`.github/workflows/run-vlab.yaml:306` passes `--eslag-leaf-groups`, `--orphan-leafs-count` and `--unbundled-servers`, but never `--multihomed-servers`, and no workflow in the repository mentions that flag.

One line fixes it for every virtual job:

```diff
           bin/hhfab vlab gen -v --mesh-links-count=$VLAB_MESH_LINKS \
             --eslag-leaf-groups=$VLAB_ESLAG_GROUPS --orphan-leafs-count=$VLAB_ORPHAN_LEAVES --unbundled-servers=$VLAB_UNBUNDLED_SERVERS \
+            --multihomed-servers=1 \
             --externals-bgp=1 --externals-static=1 --external-orphan-connections=1 \
             ${{ inputs.custom_gen_args }}
```

l2vni and l3vni both come for free with that single line, because the existing per-mode variables already differ. I ran both to confirm, on a branch carrying only that change, scoped to this test with `releasetest_regex`.

l2vni, https://github.com/githedgehog/fabricator/actions/runs/31399448182:

```
14:42:41 INF >>> eslagServers=2 unbundledServers=2 bundledServers=1 multihomedServers=1
14:52:24 INF * Running test test="HostBGP Multihoming"
14:54:23 INF PASS test="HostBGP Multihoming"
14:54:23 INF Test suite summary tests=4 passed=1 skipped=3 failed=0 duration=1m59s
```

l3vni, https://github.com/githedgehog/fabricator/actions/runs/31399457475:

```
14:42:42 INF >>> eslagServers=0 unbundledServers=2 bundledServers=1 multihomedServers=1
14:52:12 INF * Running test test="HostBGP Multihoming"
14:54:10 INF PASS test="HostBGP Multihoming"
```

So the cost is one line and about two minutes of test time per job.

Taking both modes is worth it, because they are not the same test. In l2vni the multihomed server's two links land on two leaves of the same ESLAG redundancy group, since `pkg/hhfab/vlabbuilder.go:381` only puts orphan leaves in the candidate list when there are at least two of them. In l3vni they land on two orphan leaves, and the test additionally exercises its own ESLAG skip branch when picking the two regular servers.

I left the `upgradefrom` generate line at `run-vlab.yaml:208` alone, since it runs the previously released hhfab binary.

HLAB needs a wiring change in `githedgehog/lab-ci` rather than a flag, so I have kept it out of this comment.

pkg/hhfab/rt_no_vpc_suite.go
edipascale: we do the same thing in every other failover test; the shutdown command should be nearly instantaneous (it's an administrative command on the switch); if anything, I would say that not waiting for the routes to stabilize would _create_ issues rather than mask them - e.g. the interface is now down and the route has not been updated; the fact that we never observe any flake tells me that this is not the case. we _could_ add operational status checks to assert that the port is down but it feels unjustified to me, at least given the current evidence. One thing we could do if we wanted to be more rigorous is shutdown also the other port and assert failure, unsure whether it's worth doing it but I'm open to the idea

pkg/hhfab/rt_no_vpc_suite.go
edipascale: VIPs are assigned by zebra as part of the config, there is no convergence and no need to wait for more than 1 second. The retry in other places was a defensive measure I added when I first implemented the feature as I was observing issues, until I realized that they were unrelated (bad config). I wish I had removed it, because it constantly tricks AI reviewers into wanting to add retries for this; maybe the right thing to do would be to remove it everywhere

pkg/hhfab/rt_no_vpc_suite.go
edipascale: we do this everywhere, for all tests with VPCs. It does not matter because setup-vpcs is idempotent - the first thing it does is unconditionally removing hostbgp container and calling hhnet cleanup

pau-hedgehog (changes requested):
Requesting changes on two of the inline comments above. The rest are up to you and should not hold this up.

The failover check at `pkg/hhfab/rt_no_vpc_suite.go:1518`. It shuts a port and then re-runs the same matrix, so it passes whether or not the link went down. It looks like failover coverage in the suite name and in the test report without being able to fail. In a run where the test passes, nothing in the log tells you whether the link really went down or whether the sonic-cli just accepted a command that did nothing. This one gets harder to fix once people start trusting it.

The single read of the loopback addresses at `pkg/hhfab/rt_no_vpc_suite.go:1347`. Both of the other hostBGP paths retry, this one waits one second for two addresses and gives up. There is also no log of what `getvips` returned, so if it does fail in CI you get an error and nothing to go on.

The rest is yours to take or leave: the failover switch coming out of map order, the servers left configured when the test finishes, the source address only reaching one matrix phase, and the small stuff.

The test itself works, and that is worth saying. Checking peering and isolation across two hostBGP VPCs on one multihomed server is real coverage that the auto-generated matrix cannot produce, and it passes in both VPC modes.

pkg/hhfab/rt_no_vpc_suite.go
edipascale: an "earlier suite" could belong to earlier runs, and we might change the order of suites in the future. I don't think this is worth a change

pkg/hhfab/rt_no_vpc_suite.go
edipascale: addressed

pkg/hhfab/matrix.go
edipascale: added a comment to document this

pkg/hhfab/rt_no_vpc_suite.go
edipascale: fixed

pkg/hhfab/rt_single_vpc_suite.go
edipascale: fixed

edipascale:
> > So how do we handle this? Do we create specific wirings to test this such as [githedgehog/lab#420](https://github.com/githedgehog/lab/pull/420) ? Otherwise this can't be tested as often as it should
> 
> my idea is to have the standard CI vlab job also generate a multihomed server now that they are supported by SetupVPCs, but I did not want to introduce a potentially breaking change like this right before the release

added the multihomed server to the CI as requested, since you already verified that it does not break CI

pkg/hhfab/rt_no_vpc_suite.go
edipascale: @pau-hedgehog I've added a commit where we shut down all links as a final step, and verify that no VPC on the multihomed server can be reached; it works as intended, as tested on a locally spawned VLAB. I'm not sure it is worth keeping, but I'll defer that decision to you - it does not add much to the test either way

pau-hedgehog (approved):
Thanks for the changes

Frostman (approved):

PR #1751, opened 2026-05-14
chore: bump to May 2026 flatcar

### LINKED ISSUE #1750: May 2026 Flatcar

mrbojangles3:
closes #1750

### DIFF
--- a/pkg/fab/recipe/flatcar/os_install.go
+++ b/pkg/fab/recipe/flatcar/os_install.go
@@ -187,7 +187,8 @@ func (i *OSInstal) Run(ctx context.Context) error {
 		return fmt.Errorf("partprobing: %w", err)
 	}
 
-	slog.Info("Expanding On-disk root parition", "dev", i.TargetDisk)
+	slog.Info("Expanding On-disk root partition", "dev", i.TargetDisk)
+
 	// have to delete existing partition
 	if err := i.execCmd(ctx, true, "sgdisk", "--delete=9", i.TargetDisk); err != nil {
 		return fmt.Errorf("deleting partition 9 from existing block device: %w", err)
@@ -197,11 +198,9 @@ func (i *OSInstal) Run(ctx context.Context) error {
 		return fmt.Errorf("partprobing: %w", err)
 	}
 
-	// 4857856 is the start sector start of the too small root parition
-	// not expected to change often, disk_layout is set by flatcar
 	// The typecode listed here is a UUID that flatcar uses - https://github.com/flatcar/init/blob/flatcar-master/scripts/extend-filesystems#L15
-	// Called COREOS_RESIZE, we are doing a small expand, then letting the installer exapand to the full disk size.
-	if err := i.execCmd(ctx, true, "sgdisk", "--new=9:4857856:+9G", "--typecode=9:3884dd41-8582-4404-b9a8-e9b84f2df50e", i.TargetDisk); err != nil {
+	// Called COREOS_RESIZE, expand to largest available size
+	if err := i.execCmd(ctx, true, "sgdisk", "--largest-new=9", "--typecode=9:3884dd41-8582-4404-b9a8-e9b84f2df50e", i.TargetDisk); err != nil {
 		return fmt.Errorf("creating partition 9 on existing block device: %w", err)
 	}
 
--- a/pkg/fab/recipe/node_butane.tmpl.yaml
+++ b/pkg/fab/recipe/node_butane.tmpl.yaml
@@ -82,6 +82,24 @@ storage:
           Name=dummy0
           Kind=dummy
 
+    - path: /etc/systemd/network/10-e1000-offloads.link
+      mode: 0644
+      contents:
+        inline: |
+          [Match]
+          Driver=e1000
+
+          [Link]
+          TCPSegmentationOffload=false
+          TCP6SegmentationOffload=false
+          GenericSegmentationOffload=false
+          GenericReceiveOffload=false
+          LargeReceiveOffload=false
+          ReceiveChecksumOffload=false
+          TransmitChecksumOffload=false
+          ReceiveVLANCTAGHardwareAcceleration=false
+          TransmitVLANCTAGHardwareAcceleration=false
+
     - path: /etc/systemd/network/11-dummy.network
       mode: 0644
       contents:
--- a/pkg/hhfab/vlabconfig.go
+++ b/pkg/hhfab/vlabconfig.go
@@ -198,7 +198,7 @@ type VMSize struct {
 var DefaultSizes = VMSizes{
 	Control:  VMSize{CPU: 6, RAM: 6144, Disk: 100}, // TODO 8GB RAM?
 	Switch:   VMSize{CPU: 4, RAM: 5120, Disk: 50},  // TODO 6GB RAM?
-	Server:   VMSize{CPU: 2, RAM: 768, Disk: 10},   // TODO 1GB RAM?
+	Server:   VMSize{CPU: 2, RAM: 768, Disk: 14},   // TODO 1GB RAM?
 	Gateway:  VMSize{CPU: 8, RAM: 6144, Disk: 50},  // TODO proper min size
 	External: VMSize{CPU: 4, RAM: 5120, Disk: 50},  // TODO 6GB RAM?
 }
--- a/pkg/hhfab/vlabrunner.go
+++ b/pkg/hhfab/vlabrunner.go
@@ -394,11 +394,7 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 				if err := execCmd(ctx, false, vmDir,
 					VLABCmdQemuImg, []string{"resize", VLABOSImageFile, fmt.Sprintf("%dG", vm.Size.Disk)},
 					"vm", vm.Name); err != nil {
-					if err := execCmd(ctx, false, vmDir,
-						VLABCmdQemuImg, []string{"resize", "--shrink", VLABOSImageFile, fmt.Sprintf("%dG", vm.Size.Disk)},
-						"vm", vm.Name); err != nil {
-						return fmt.Errorf("resizing os image: %w", err)
-					}
+					return fmt.Errorf("resizing os image: %w", err)
 				}
 			}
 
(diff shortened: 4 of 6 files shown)


Frostman (dismissed):

mrbojangles3:
Somehow this version of flatcar/kernel/systemd the nic names are:
```
core@control-1 ~ $ ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: enp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 0c:20:12:fe:00:00 brd ff:ff:ff:ff:ff:ff
    altname enx0c2012fe0000
3: ens1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 0c:20:12:fe:00:01 brd ff:ff:ff:ff:ff:ff
    altname enp2s1
    altname enx0c2012fe0001
4: dummy0: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether 06:66:b3:e6:12:19 brd ff:ff:ff:ff:ff:ff
```
I am looking for a way to get the previous naming as the primary name.

pkg/fab/recipe/control_butane.tmpl.yaml
```diff
@@ -63,6 +63,7 @@ storage:
           Driver=e1000
 
           [Link]
+          NamePolicy=path
```
mrbojangles3: This line is what forces the old naming policy for our nics in this case. This file is getting applied and we needed to have this name. Why now and not when it went in? I don't know.

pkg/hhfab/vlabrunner.go
```diff
@@ -393,7 +393,11 @@ func (c *Config) VLABRun(ctx context.Context, vlab *VLAB, opts VLABRunOpts) erro
 				if err := execCmd(ctx, false, vmDir,
 					VLABCmdQemuImg, []string{"resize", VLABOSImageFile, fmt.Sprintf("%dG", vm.Size.Disk)},
 					"vm", vm.Name); err != nil {
-					return fmt.Errorf("resizing os image: %w", err)
+					if err := execCmd(ctx, false, vmDir,
```
mrbojangles3: This is meant to be a retry, assuming the resize is meant to grow the first `if` will succeed. `qemu-img` will fail if you are shrinking a disk. So assuming we have failed then we need to shrink the disk. 

The other way to fix this would be to read the existing size, and adjust the args accordingly. but this isn't a standard `os.exec` where we can read the output so instead of writing all the code to make that happen, we just try again.

pkg/fab/recipe/flatcar/os_install.go
```diff
@@ -197,11 +198,9 @@ func (i *OSInstal) Run(ctx context.Context) error {
 		return fmt.Errorf("partprobing: %w", err)
 	}
 
-	// 4857856 is the start sector start of the too small root parition
-	// not expected to change often, disk_layout is set by flatcar
 	// The typecode listed here is a UUID that flatcar uses - https://github.com/flatcar/init/blob/flatcar-master/scripts/extend-filesystems#L15
 	// Called COREOS_RESIZE, we are doing a small expand, then letting the installer exapand to the full disk size.
-	if err := i.execCmd(ctx, true, "sgdisk", "--new=9:4857856:+9G", "--typecode=9:3884dd41-8582-4404-b9a8-e9b84f2df50e", i.TargetDisk); err != nil {
+	if err := i.execCmd(ctx, true, "sgdisk", "--largest-new=9", "--typecode=9:3884dd41-8582-4404-b9a8-e9b84f2df50e", i.TargetDisk); err != nil {
```
mrbojangles3, line 203: When the disk sizes changes in the future, we won't have the hard-coded sectors. Unfortunately I didn't see the option the first time.

pkg/hhfab/vlabconfig.go
```diff
@@ -198,7 +198,7 @@ type VMSize struct {
 var DefaultSizes = VMSizes{
 	Control:  VMSize{CPU: 6, RAM: 6144, Disk: 100}, // TODO 8GB RAM?
 	Switch:   VMSize{CPU: 4, RAM: 5120, Disk: 50},  // TODO 6GB RAM?
-	Server:   VMSize{CPU: 2, RAM: 768, Disk: 10},   // TODO 1GB RAM?
+	Server:   VMSize{CPU: 2, RAM: 768, Disk: 14},   // TODO 1GB RAM?
```
mrbojangles3, line 201: The flatcar_qemu image that we use to boot up vlab is a little over 12.3GiB, if we resize the image down before we shrink the file system partitions etc, we will have problems. Bumping the size avoids that problem.

pkg/fab/recipe/node_butane.tmpl.yaml
```diff
@@ -114,6 +114,18 @@ storage:
...
+      mode: 0644
+      contents:
+        inline: |
+          [Match]
+          OriginalName=*
+
+          [Link]
+          NamePolicy= keep path
+          AlternativeNamesPolicy=database onboard slot path
+          MACAddressPolicy=persistent
```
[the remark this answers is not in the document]
mrbojangles3: will change

pkg/fab/recipe/install_upgrade.go
```diff
@@ -280,6 +282,35 @@ func setupTimesync(ctx context.Context, controlVIP string) error {
 	return nil
 }
 
+func installNetworkFile(_ context.Context) error {
+	path := filepath.Join(LinkPath, LinkFileName)
+	switch _, err := os.Stat(path); { //nolint: whitespace
```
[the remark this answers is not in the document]
mrbojangles3: our linter is set to not like the true idiom. not changing

pkg/fab/recipe/install_upgrade.go
```diff
@@ -280,6 +282,35 @@ func setupTimesync(ctx context.Context, controlVIP string) error {
...
+	case errors.Is(err, os.ErrNotExist):
+		slog.Info("Installing 98-e1000.link")
+		const file = `[Match]
+Driver=e1000
+
+[Link]
+NamePolicy=keep path
+AlternativeNamesPolicy=database onboard slot path
+MACAddressPolicy=persistent
+`
```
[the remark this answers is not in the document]
mrbojangles3: I don't see this in my editor. Also this is not a butane file. Not changing.

pkg/fab/recipe/flatcar/os_install.go
```diff
@@ -197,11 +197,9 @@ func (i *OSInstal) Run(ctx context.Context) error {
 		return fmt.Errorf("partprobing: %w", err)
 	}
 
-	// 4857856 is the start sector start of the too small root parition
-	// not expected to change often, disk_layout is set by flatcar
 	// The typecode listed here is a UUID that flatcar uses - https://github.com/flatcar/init/blob/flatcar-master/scripts/extend-filesystems#L15
 	// Called COREOS_RESIZE, we are doing a small expand, then letting the installer exapand to the full disk size.
```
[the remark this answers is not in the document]
mrbojangles3: Will change

pkg/fab/recipe/control_butane.tmpl.yaml
Frostman: Don't we need 

```
          AlternativeNamesPolicy=database onboard slot path
          MACAddressPolicy=persistent
```

here as well?

pkg/fab/recipe/install_upgrade.go
```diff
@@ -280,6 +282,52 @@ func setupTimesync(ctx context.Context, controlVIP string) error {
 	return nil
 }
 
+func installNetworkFile(_ context.Context) error {
```
Frostman: Looking at the butane files, only `10-e1000-offloads.link` is being installed on the actual nodes; the `98-e1000.link` is only used in a live flatcar installer. It seems like we don't need it on the nodes as things are being handled by `10-e1000-offloads.link`.

pkg/fab/recipe/install_upgrade.go
```diff
@@ -280,6 +282,52 @@ func setupTimesync(ctx context.Context, controlVIP string) error {
...
+	case err != nil:
+		return fmt.Errorf("error with stat 98-e1000.link: %w", err)
+	default:
+		slog.Info("98-e1000.link exists, skipping")
+
+		return nil
+	}
+}
+
+func editOffloadFile(_ context.Context) error {
```
Frostman: I think we should just unconditionally override the file so that it to 100% match what we have in the butane. Current approach can create issue if somebody edited a file or commended the NamePolicy line. Assuming other comment about aligning the files used and their contents between control node and other nodes - you can just name if enforceE1000Workaround or smth like this and call for both control_upgrade and node_upgrade paths while for install it's handled by the butane/ignition.

pkg/fab/recipe/node_butane.tmpl.yaml
```diff
@@ -114,6 +114,18 @@ storage:
           EmitLLDP=yes
           ConfigureWithoutCarrier=yes
 
+    - path: /etc/systemd/network/98-e1000.link
```
Frostman: you're using `10-e1000-offloads.link` on the control node, let's use the same file for nodes too and please make a separate commit in this PR to add same offloads trick here like on control node, I think we just missed it before. So it that case files would be 100% aligned.

Frostman (changes requested):
Few comments and waiting for final confirmation after test of upgrading control node from 26.02 on real hardware.

pkg/fab/recipe/control_butane.tmpl.yaml
mrbojangles3: Not specifically, Those lines were when I was putting in a `.link` file that was going to apply to multiple interfaces. Since we are limiting the change to e1000 we don't need these lines.

Frostman:
@coderabbitai review

pau-hedgehog:
For the record, we should add [ci:+hlab](https://github.com/githedgehog/fabricator/issues?q=state%3Aopen%20label%3Aci%3A%2Bhlab) in all flatcar bump PRs as it can fail as it's the current case

pkg/fab/recipe/control_butane.tmpl.yaml
mrbojangles3: Still need to remove the path from this line

mrbojangles3:
@Frostman This PR is going to pass this round of CI and will finally be able to get merged

Frostman (approved):

PR #1759, opened 2026-05-19
test(release): add port-channel fallback test

pau-hedgehog:
## Summary

Adds an `ESLAG Fallback` release test in the Single VPC Suite. The test picks an ESLAG connection in the active wiring, enables `fallback: true` on it (with a revert), and for each link in turn brings up exactly that one slave NIC with a VLAN sub-interface via `hhnet` and asserts a DHCP lease is received. Every leg must succeed. Tracks githedgehog/fabric#1467.

The assertion is data-plane only. The failure message names the slave NIC, the leaf, and the connection so the failing leaf is identifiable from a CI log. Teardown rebuilds the bond via `hhnet cleanup` plus the standard netconf command, mirroring `setupServerNetwork`, so subsequent tests inherit a clean server.

Registered next to the existing `ESLAG Failover` test.

## Test plan
Before fix is bumped:
  - [x] Failing leg returns 0/3 leases on a wiring where fallback is not uniformly programmed; suite FAILs with `slave=<...> leaf=<...> connection=<...>`.

When https://github.com/githedgehog/fabric/pull/1466 is bumped into fabricator:
  - [x] `tmp:` commits are dropped and remaining ones squashed
  - [x] All checks pass; release tests pass on HLAB (test skips on VLAB).

### DIFF
--- a/pkg/hhfab/rt_eslag_fallback.go
+++ b/pkg/hhfab/rt_eslag_fallback.go
@@ -1,0 +1,237 @@
+// Copyright 2026 Hedgehog
+// SPDX-License-Identifier: Apache-2.0
+
+package hhfab
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/netip"
+	"strings"
+
+	vpcapi "go.githedgehog.com/fabric/api/vpc/v1beta1"
+	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
+	"go.githedgehog.com/fabricator/pkg/util/sshutil"
+	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	eslagPxeDHCPTimeoutSec  = 20
+	eslagPxeAttemptsPerLeg  = 3
+	eslagPxeAttemptsPerLegX = 5
+)
+
+func eslagFallbackTest(ctx context.Context, testCtx *VPCPeeringTestCtx) (bool, []RevertFunc, error) {
+	if testCtx.setupOpts.VPCMode == vpcapi.VPCModeL3VNI {
+		return true, nil, fmt.Errorf("L3VNI mode is not compatible with ESLAG") //nolint:goerr113
+	}
+
+	conns := &wiringapi.ConnectionList{}
+	if err := testCtx.kube.List(ctx, conns, kclient.MatchingLabels{wiringapi.LabelConnectionType: wiringapi.ConnectionTypeESLAG}); err != nil {
+		return false, nil, fmt.Errorf("listing ESLAG connections: %w", err)
+	}
+
+	candidates := make([]*wiringapi.Connection, 0)
+	for i := range conns.Items {
+		c := &conns.Items[i]
+		if c.Spec.ESLAG != nil && len(c.Spec.ESLAG.Links) >= 2 {
+			candidates = append(candidates, c)
+		}
+	}
+	if len(candidates) == 0 {
+		return true, nil, errNoEslags
+	}
+
+	if !testCtx.extended {
+		candidates = candidates[:1]
+	}
+
+	attempts := eslagPxeAttemptsPerLeg
+	if testCtx.extended {
+		attempts = eslagPxeAttemptsPerLegX
+	}
+
+	reverts := make([]RevertFunc, 0)
+	ranAtLeastOne := false
+	for _, conn := range candidates {
+		vlan, server, err := findESLAGAttachedSubnet(ctx, testCtx.kube, conn)
+		if err != nil {
+			return false, reverts, fmt.Errorf("finding VPC attachment for connection %s: %w", conn.Name, err)
+		}
+		if vlan == 0 {
+			slog.Info("ESLAG connection has no VPC attachment in the active suite, skipping", "connection", conn.Name)
+
+			continue
+		}
+
+		if err := enableESLAGFallback(ctx, testCtx, conn.Name, &reverts); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,263 @@
...
+	candidates := make([]*wiringapi.Connection, 0)
+	for i := range conns.Items {
+		c := &conns.Items[i]
+		if c.Spec.ESLAG != nil && len(c.Spec.ESLAG.Links) >= 2 {
+			candidates = append(candidates, c)
+		}
+	}
+	if len(candidates) == 0 {
+		return true, nil, errNoEslags
+	}
```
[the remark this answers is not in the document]
pau-hedgehog: By design; PR description updated to match.

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,263 @@
...
+		netconfCmd, err := GetServerNetconfCmd(conn, ServerNetconfOpts{
+			VLAN:       vlan,
+			HashPolicy: testCtx.setupOpts.HashPolicy,
+		})
+		if err != nil {
+			return false, reverts, fmt.Errorf("building netconf command for %s: %w", server, err)
+		}
+		reverts = append(reverts, func(ctx context.Context) error {
+			return restoreServerBond(ctx, ssh, netconfCmd)
+		})
```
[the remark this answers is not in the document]
pau-hedgehog: Per-iteration `:=` locals; no stale capture.

pkg/hhfab/rt_eslag_fallback.go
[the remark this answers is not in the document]
pau-hedgehog: ESLAG spec allows 2-4 (API MinItems=2).

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,237 @@
...
+
+	if err := WaitReady(ctx, testCtx.kube, testCtx.wrOpts); err != nil {
+		return fmt.Errorf("waiting for switches after enabling fallback on %s: %w", name, err)
+	}
+
+	return nil
+}
+
+// findESLAGAttachedSubnet returns vlan=0 if the connection has no attachment.
+func findESLAGAttachedSubnet(ctx context.Context, kube kclient.Client, conn *wiringapi.Connection) (uint16, string, error) {
```
edipascale, line 171: nit: rename to clarify that it returns the VLAN, not the subnet prefix

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,237 @@
...
+	_, serverNames, _, _, err := conn.Spec.Endpoints() //nolint:dogsled
+	if err != nil || len(serverNames) != 1 {
+		return 0, "", fmt.Errorf("connection %s does not have a single server endpoint", conn.Name) //nolint:goerr113
+	}
+	server := serverNames[0]
+
+	attaches := &vpcapi.VPCAttachmentList{}
+	if err := kube.List(ctx, attaches); err != nil {
```
edipascale, line 179: you could filter by connection name:
```
if err := kube.List(ctx, attaches, kclient.MatchingLabels{wiringapi.LabelConnection: conn.Name}); err != nil {
```

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,237 @@
...
+	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
+	"go.githedgehog.com/fabricator/pkg/util/sshutil"
+	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	eslagPxeDHCPTimeoutSec  = 20
+	eslagPxeAttemptsPerLeg  = 3
+	eslagPxeAttemptsPerLegX = 5
```
edipascale, line 23: question, why do we need more attempts in the extended case?

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,237 @@
...
+					ok++
+				}
+			}
+			slog.Info("Port-channel fallback leg result", "server", server, "slave", slave, "leaf", leafForSlave[slave], "ok", ok, "attempts", attempts)
+			if ok == 0 {
+				failures = append(failures, fmt.Sprintf("slave=%s leaf=%s connection=%s", slave, leafForSlave[slave], conn.Name))
+			}
+		}
+
+		ranAtLeastOne = true
```
edipascale, line 114: nit: this should probably be called `ranAtLeastOnce` and set to true inside the for loop with the pxeAttempt, we're not checking anywhere that `len(slaves) > 0` even though in practice this should never happen

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -0,0 +1,237 @@
...
+			continue
+		}
+
+		return subnet.VLAN, server, nil
+	}
+
+	return 0, server, nil
+}
+
+// pxeAttempt bounds hhnet's 5-minute DHCP wait with timeout(1) so a
```
edipascale, line 201: nit: timeout(1) in the comment is misleading, the timeout is a parameter and its constant value is 3

edipascale (approved):
thanks @pau-hedgehog, the test looks mostly good, just a few comments / nits

Frostman (approved):
@pau-hedgehog I'm merging it, but there are good comments from @edipascale to take care of

PR #1761, opened 2026-05-21
feat: test-connectivity refactor with matrix

edipascale:
replace a number of ad-hoc testing function with a declarative, matrix-based test. Reachability checks that are currently supported by Fabric are preserved as is; additional checks are manually overlayed on the matrix, e.g. for NAT tests.

### DIFF
--- a/pkg/hhfab/endpoints.go
+++ b/pkg/hhfab/endpoints.go
@@ -1,0 +1,366 @@
+// Copyright 2026 Hedgehog
+// SPDX-License-Identifier: Apache-2.0
+
+package hhfab
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/netip"
+	"slices"
+	"strings"
+	"sync"
+
+	vpcapi "go.githedgehog.com/fabric/api/vpc/v1beta1"
+	wiringapi "go.githedgehog.com/fabric/api/wiring/v1beta1"
+	"go.githedgehog.com/fabricator/pkg/util/sshutil"
+	"golang.org/x/sync/errgroup"
+	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SSHResolver returns the SSH config for a server. The collector and per-
+// server refresh helpers take one of these instead of a pre-built
+// map[string]*sshutil.Config so callers can share whatever SSH plumbing
+// they already have (SetupVPCs builds a map keyed by VM name; rt_* tests
+// resolve lazily through testCtx.getSSH).
+type SSHResolver func(server string) (*sshutil.Config, error)
+
+// SSHResolverFromMap adapts a pre-built map into an SSHResolver.
+func SSHResolverFromMap(m map[string]*sshutil.Config) SSHResolver {
+	return func(server string) (*sshutil.Config, error) {
+		cfg, ok := m[server]
+		if !ok {
+			return nil, fmt.Errorf("no ssh config for server %q", server) //nolint:goerr113
+		}
+
+		return cfg, nil
+	}
+}
+
+// discoveredIP pairs a server-side interface name with the address found
+// on it. The interface tells us whether the address is a hostBGP /32 VIP
+// (on `lo`) or a regular subnet address (on the bond/VLAN interface).
+type discoveredIP struct {
+	iface  string
+	prefix netip.Prefix
+}
+
+// discoverServerIPs returns every eligible IPv4 address configured on the
+// server, paired with the interface it lives on. The management interface
+// (enp2s0), the docker bridge (docker0), and the loopback 127.0.0.1/8 entry
+// are skipped; other lo addresses (hostBGP /32 VIPs) are kept.
+func discoverServerIPs(ctx context.Context, sshCfg *sshutil.Config, server string) ([]discoveredIP, error) {
+	stdout, stderr, err := sshCfg.Run(ctx, "ip -o -4 addr show | awk '{print $2, $4}'")
+	if err != nil {
+		return nil, fmt.Errorf("running ip addr show on %s: %w: %s", server, err, stderr)
+	}
+
+	lines := strings.Split(strings.TrimSpace(stdout), "\n")
+	out := make([]discoveredIP, 0, len(lines))
+	for _, line := range lines {
+		fields := strings.Fields(line)
+		if len(fields) != 2 {
+			continue
+		}
+		if (fields[0] == "lo" && fields[1] == "127.0.0.1/8") || fields[0] == "enp2s0" || fields[0] == "docker0" {
+			continue
+		}
+		prefix, err := netip.ParsePrefix(fields[1])
+		if err != nil {
+			return nil, fmt.Errorf("parsing %q on %s: %w", fields[1], server, err)
+		}
+		out = append(out, discoveredIP{iface: fields[0], prefix: prefix})
+	}
+
+	return out, nil
+}
+
+// serverAttachment is the (vpc, subnet) information the collector resolves
+// from a VPCAttachment + its referenced VPC CRD.
+type serverAttachment struct {
+	vpcName    string
+	subnetName string
+	subnetCIDR netip.Prefix
+	hostBGP    bool
+	attachName string // for diagnostics
+}
+
+// CollectServerEndpoints observes the live cluster and produces one
+// *Endpoint per (server, vpc, subnet) attachment for each server in the
+// `servers` filter (nil → all servers attached to at least one VPC).
+//
+// Algorithm: list VPCAttachments → group by server (resolved via each
+// attachment's Connection); for each candidate server, SSH it and read all
+// IPv4 addresses; match each address to one of its attachments by CIDR
+// containment (narrowest prefix wins on ties). HostBGP is derived from
+// VPCSubnet.HostBGP. An attachment with no matching configured address is
+// dropped with a warning; a server with no configured addresses contributes
+// nothing — this matches today's SetupVPCs behavior, where ESLAG servers
+// in L3VNI mode are skipped before hhnet runs and therefore have no IPs.
+//
+// Returns an error only when SSH itself fails on a queried server or when a
+// VPCAttachment cannot be resolved to a Connection.
+func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHResolver, servers []string) ([]*Endpoint, error) {
+	attaches := &vpcapi.VPCAttachmentList{}
+	if err := kube.List(ctx, attaches); err != nil {
+		return nil, fmt.Errorf("listing VPCAttachments: %w", err)
+	}
+
+	connCache := map[string]*wiringapi.Connection{}
+	getConn := func(name string) (*wiringapi.Connection, error) {
+		if c, ok := connCache[name]; ok {
+			return c, nil
+		}
+		c := &wiringapi.Connection{}
+		if err := kube.Get(ctx, kclient.ObjectKey{Name: name, Namespace: kmetav1.NamespaceDefault}, c); err != nil {
+			return nil, fmt.Errorf("getting connection %q: %w", name, err)
+		}
+		connCache[name] = c
+
+		return c, nil
+	}
+
+	vpcCache := map[string]*vpcapi.VPC{}
+	getVPC := func(name string) (*vpcapi.VPC, error) {
+		if v, ok := vpcCache[name]; ok {
+			return v, nil
+		}
+		v := &vpcapi.VPC{}
+		if err := kube.Get(ctx, kclient.ObjectKey{Name: name, Namespace: kmetav1.NamespaceDefault}, v); err != nil {
+			return nil, fmt.Errorf("getting VPC %q: %w", name, err)
+		}
+		vpcCache[name] = v
+
+		return v, nil
+	}
+
+	want := map[string]bool{}
+	for _, s := range servers {
+		want[s] = true
+	}
+
+	serverAttachments := map[string][]serverAttachment{}
+	for _, attach := range attaches.Items {
+		conn, err := getConn(attach.Spec.Connection)
+		if err != nil {
+			return nil, fmt.Errorf("resolving attachment %q: %w", attach.Name, err)
+		}
+		_, srvs, _, _, err := conn.Spec.Endpoints()
+		if err != nil {
+			return nil, fmt.Errorf("getting endpoints of connection %q: %w", conn.Name, err)
+		}
+		if len(srvs) != 1 {
+			// VPCAttachments only reference server-facing connections; if a
+			// connection has no server endpoint we skip it as a malformed
+			// attachment rather than erroring out, since fabric webhooks
+			// already gate this.
+			continue
+		}
+		serverName := srvs[0]
+		if len(want) > 0 && !want[serverName] {
+			continue
+		}
+
+		vpc, err := getVPC(attach.Spec.VPCName())
+		if err != nil {
+			return nil, fmt.Errorf("resolving attachment %q: %w", attach.Name, err)
+		}
+		subnetName := attach.Spec.SubnetName()
+		subnet, ok := vpc.Spec.Subnets[subnetName]
+		if !ok {
+			return nil, fmt.Errorf("attachment %q references missing subnet %s/%s", attach.Name, vpc.Name, subnetName) //nolint:goerr113
+		}
+		cidr, err := netip.ParsePrefix(subnet.Subnet)
+		if err != nil {
+			return nil, fmt.Errorf("parsing VPC %s/%s subnet CIDR %q: %w", vpc.Name, subnetName, subnet.Subnet, err)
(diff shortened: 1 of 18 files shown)


pkg/hhfab/testing.go
```diff
@@ -2311,83 +2520,54 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
...
-							"expected", expectedReachable.Reachable,
-						}
-						if expectedReachable.Reachable {
-							logArgs = append(logArgs, "reason", expectedReachable.Reason)
-							if expectedReachable.Peering != "" {
-								logArgs = append(logArgs, "peering", expectedReachable.Peering)
-							}
-						}
-						slog.Debug("Checking connectivity", logArgs...)
+					expectedReachable, err := IsServerReachable(ctx, kube, serverA, serverB, c.Fab.Spec.Config.Gateway.Enable)
```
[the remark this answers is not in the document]
edipascale: false

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,784 @@
...
+	sshByServer    map[string]*sshutil.Config
+	toolboxMutexes map[string]*sync.Mutex
+	pings          *semaphore.Weighted
+	iperfs         *semaphore.Weighted
+	curls          *semaphore.Weighted
+	inSources      func(string) bool
+	inDestinations func(string) bool
+	wg             *sync.WaitGroup
+	errChan        chan<- error
+}
```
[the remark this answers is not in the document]
edipascale: false

edipascale:
@pau-hedgehog I plan on doing another pass to remove some of the verbose comments and simplify the code in a few places, but this should be ready to be reviewed, at least in terms of the approach used

pkg/hhfab/testing.go
```diff
@@ -1829,12 +1845,118 @@ func (c *Config) SetupPeerings(ctx context.Context, vlab *VLAB, opts SetupPeerin
...
+					if r.Reachable {
+						m.Add(ConnectivityExpectation{
+							Pair:    EndpointPair{Source: src, Destination: dst},
+							Verdict: VerdictAllow,
+							Reason:  r.Reason,
+							Peering: r.Peering,
+						})
+
+						break
+					}
```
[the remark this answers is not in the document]
edipascale: Your assessment is correct but this is intentional; we currently rely on manual entries in the matrix for multi-homed servers and/or NAT, as a stop-gap solution to be able to test scenarios using those features until we have reliable reachibility helpers supporting all use cases. If this design is approved, those helpers will be added over time in separate PRs

pkg/hhfab/rt_nat_tests.go
```diff
@@ -720,6 +614,20 @@ func (testCtx *VPCPeeringTestCtx) gatewayPeeringOverlapNATTest(ctx context.Conte
...
+	// server that is, by then, back in donorVPC. We hold the original
+	// *Endpoint pointers — they are still valid after ReplaceServerEndpoints
+	// removes them from AllEndpoints, and re-installing them preserves the
+	// IPs captured at suite setup.
+	var origEndpoints []*Endpoint
+	for _, ep := range matrix.AllEndpoints {
+		if ep.Server != nil && ep.Server.Name == targetServer {
+			origEndpoints = append(origEndpoints, ep)
+		}
+	}
```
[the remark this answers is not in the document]
edipascale: addressed; I also removed the server name from the map value as that was now redundant

pkg/hhfab/rt_nat_tests.go
```diff
@@ -10,24 +10,88 @@ import (
...
+// Returns (skip=true) when fewer than two VPCs are available so the suite
+// can mark the case as skipped.
+func (testCtx *VPCPeeringTestCtx) runNATTest(ctx context.Context, matrix *ConnectivityMatrix, spec natTestSpec) (bool, []RevertFunc, error) {
+	vpcs := &vpcapi.VPCList{}
+	if err := testCtx.kube.List(ctx, vpcs); err != nil {
+		return false, nil, fmt.Errorf("listing VPCs: %w", err)
+	}
+	if len(vpcs.Items) < 2 {
+		return true, nil, fmt.Errorf("not enough VPCs for %s test", spec.Name) //nolint:goerr113
+	}
```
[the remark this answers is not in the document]
edipascale: that's incorrect, error messages are used to convey the reason for skipping the test if skip=true

pkg/hhfab/testing.go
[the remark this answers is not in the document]
edipascale: no need, I'll open an issue to track this issue if this PR is approved and merged

pau-hedgehog:
@edipascale I rebased to see how stable this is while reviewing and I'm getting consistent failures in:

https://github.com/githedgehog/fabricator/actions/runs/29049747113/job/86267142288

edipascale:
> @edipascale I rebased to see how stable this is while reviewing and I'm getting consistent failures in:
> 
> https://github.com/githedgehog/fabricator/actions/runs/29049747113/job/86267142288

thanks @pau-hedgehog, I'll have a look to see what changed, as this used to pass steadily. However the main thing that is needed imho is feedback on whether this approach makes sense. Is it an improvement over the current state? Does it allow us to test what we want to test? Are there gaps that we should cover?

pau-hedgehog:
Filed the DHCP static lease / stale connectivity matrix issue we found while triaging this branch's failures separately: #1876

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+				Expected: expected,
+				Bidir:    bidir,
+				Pings:    deps.pings,
+				Iperfs:   deps.iperfs,
+			}
+			deps.wg.Go(func() {
+				for _, e := range runPingIperfPair(ctx, opts, args) {
+					deps.errChan <- e
+				}
+			})
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+					logArgs = append(logArgs, "peering", expected.Peering)
+				}
+			}
+			slog.Debug("Checking external connectivity", logArgs...)
+
+			if ce := checkCurl(ctx, opts, deps.curls, name, ssh, "1.0.0.1", expected.Reachable); ce != nil {
+				deps.errChan <- ce
+			}
+		})
+	}
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+			}
+		}
+	}
+	for key, val := range extTargets {
+		deps.wg.Go(func() {
+			if ie := runMatrixIperfPortForward(ctx, opts, deps.iperfs, key.from, deps.sshByServer[key.from], key.ip, key.port, val); ie != nil {
+				deps.errChan <- ie
+			}
+		})
+	}
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/endpoints.go
```diff
@@ -0,0 +1,366 @@
...
+	for name := range serverAttachments {
+		names = append(names, name)
+	}
+	slices.Sort(names)
+	for _, name := range names {
+		eg.Go(func() error {
+			cfg, err := ssh(name)
+			if err != nil {
+				return fmt.Errorf("getting ssh config for %s: %w", name, err)
+			}
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+				}
+				expected := reachabilityFromExpectation(e)
+				target := e.NAT.DestinationIP
+				port := e.NAT.DestinationPort
+				fromName := src.Server.Name
+				deps.wg.Go(func() {
+					if ie := runMatrixIperfPortForward(ctx, opts, deps.iperfs, fromName, deps.sshByServer[fromName], target, port, expected); ie != nil {
+						deps.errChan <- ie
+					}
+				})
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/testing.go
```diff
@@ -2219,55 +2418,28 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
 			ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 			defer cancel()
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/testing.go
```diff
@@ -2325,73 +2497,53 @@ func (c *Config) TestConnectivity(ctx context.Context, vlab *VLAB, opts TestConn
...
-							"expected", expectedReachable.Reachable,
-						}
-						if expectedReachable.Reachable {
-							logArgs = append(logArgs, "reason", expectedReachable.Reason)
-							if expectedReachable.Peering != "" {
-								logArgs = append(logArgs, "peering", expectedReachable.Peering)
-							}
-						}
-						slog.Debug("Checking connectivity", logArgs...)
+					expectedReachable, err := IsServerReachable(ctx, kube, serverA, serverB, c.Fab.Spec.Config.Gateway.Enable)
```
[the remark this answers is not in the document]
edipascale: not true since Go 1.22

pkg/hhfab/testing.go
```diff
@@ -1804,12 +1820,118 @@ func (c *Config) SetupPeerings(ctx context.Context, vlab *VLAB, opts SetupPeerin
...
+				continue
+			}
+			switch {
+			case src.Server != nil && dst.Server != nil:
+				if src.Server.Name == dst.Server.Name {
+					continue
+				}
+				r, err := IsServerReachable(ctx, kube, src.Server.Name, dst.Server.Name, gatewayEnabled)
+				if err != nil {
+					if errors.Is(err, reachCheckUnsupported) {
```
pau-hedgehog, line 1876: This skip writes no entry, and Lookup treats missing pairs as Deny, so a pair populate can't evaluate silently becomes "expected unreachable". 

On master the same condition fails the run loudly. Same pattern: discovery drops endpoints on
warning, and noSetup returns an empty non-nil matrix (rt_base.go:425). All of these pass green while testing nothing.

Can we check the matrix after populate/overlay before leaning on Deny verdicts?

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+		for _, dst := range matrix.AllEndpoints {
+			if dst.External == nil {
+				continue
+			}
+			e := matrix.Lookup(src, dst, ProtoPort{})
+			if e.Verdict != VerdictAllow {
+				continue
+			}
+			if e.NAT != nil && !e.NAT.SourcePool.IsValid() {
```
pau-hedgehog, line 504: SourcePool decides whether to curl, but the observed source address is never asserted. Same on master, so not blocking, just for follow-up

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+		if !deps.inSources(src.Server.Name) {
+			continue
+		}
+		for _, dst := range matrix.AllEndpoints {
+			if src == dst {
+				continue
+			}
+			e := matrix.Lookup(src, dst, ProtoPort{})
+			if e.Verdict != VerdictAllow || e.NAT == nil {
```
pau-hedgehog, line 559: Deny port-forward entries are never probed, so negative coverage doesn't reach port-forwards yet. The Deny entries already keep their NAT details, so the data is there when the runner is ready. Worth a follow-up

pkg/hhfab/rt_utils.go
```diff
@@ -302,7 +302,14 @@ func shutDownLinkAndTest(ctx context.Context, testCtx *VPCPeeringTestCtx, link w
 	slog.Debug("Waiting 5 seconds")
 	time.Sleep(5 * time.Second)
 
-	return DoVLABTestConnectivity(ctx, testCtx.vlabCfg.WorkDir, testCtx.vlabCfg.CacheDir, testCtx.tcOpts)
+	var connCheckErr error
+	if matrix == nil {
+		connCheckErr = DoVLABTestConnectivity(ctx, testCtx.vlabCfg.WorkDir, testCtx.vlabCfg.CacheDir, testCtx.tcOpts)
+	} else {
+		connCheckErr = DoVLABTestConnectivityWithMatrix(ctx, testCtx.vlabCfg.WorkDir, testCtx.vlabCfg.CacheDir, testCtx.tcOpts, matrix)
```
pau-hedgehog, line 309: Spine, gateway and mesh failover receive the matrix but still call the legacy path, while link failover already uses it here. Do you plan to migrate them later?

pau-hedgehog (commented):
Sorry it took me to long. Working through the rest of my review

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+			// is expected to target.
+			toIP := dst.Server.IP
+			if entry.NAT != nil && entry.NAT.DestinationIP.IsValid() {
+				toIP = entry.NAT.DestinationIP
+			}
+			if !toIP.IsValid() {
+				return fmt.Errorf("matrix entry %s→%s (vpc %s/%s) has no valid target IP", src.Server.Name, dst.Server.Name, dst.Server.VPC, dst.Server.Subnet) //nolint:goerr113
+			}
+
+			expected := reachabilityFromExpectation(entry)
```
pau-hedgehog, line 437: When a probe fails, the error says what was expected but not why the matrix expected it. The entry knows (Reason, Peering) but that's dropped here and only shows at debug level. Including it in the failure message would make errors easier to debug

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+	sshConfigs, _, cacheCancel, err := c.prepareConnectivityTest(ctx, vlab, &opts)
+	if err != nil {
+		return err
+	}
+	defer cacheCancel()
+
+	// Resolve the SSH config and a toolbox mutex for every unique server
+	// name referenced by the matrix.
+	sshByServer := map[string]*sshutil.Config{}
+	toolboxMutexes := map[string]*sync.Mutex{}
```
pau-hedgehog, line 630: nit: looks unused?

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+			if existing.NAT != nil {
+				nat = *existing.NAT
+			}
+			if err := mutator(src, dst, &nat); err != nil {
+				return err
+			}
+			matrix.Add(ConnectivityExpectation{
+				Pair:    EndpointPair{Source: src, Destination: dst},
+				Verdict: VerdictAllow,
+				Reason:  ReachabilityReasonGatewayPeering,
```
pau-hedgehog, line 222: nit: hardcoding reason to GatewayPeering

pau-hedgehog (dismissed):
This was huge. It improves our coverage significantly, especially the NAT tests.
There are some suggested follow-ups in the comments, the matrix guard being the main one.
Next thing would be figuring out migration next steps.

pkg/hhfab/testing.go
pau-hedgehog: A golden-file test pinning populate's output for a fixed topology could catch most of this class too

edipascale: I've attempted to address most of these comments, with this being the most prominent one, in https://github.com/githedgehog/fabricator/pull/1910
Specifically for this, the idea is to have a list of "dropped endpoints" which have IPs that do not match any subnet or attachments with no matching IPs. These are part of the matrix and will fail a validate call that runs as the first step of the test connectivity. Nodes with no IPs are still silently glossed over to be compatible with the eslag l3vni issue/workaround

pkg/hhfab/matrix.go
edipascale: agreed to leave it for a follow-up

pkg/hhfab/matrix.go
edipascale: There should now be infra to handle this, although I did not extend the existing tests to make use of it yet

pkg/hhfab/rt_utils.go
edipascale: migrated in the branch mentioned above

pkg/hhfab/matrix.go
edipascale: a "why" including reason and peering (or whichever of the two are available if we only have one) has been added

pkg/hhfab/matrix.go
edipascale: removed

pkg/hhfab/matrix.go
edipascale: an existing reason is now preserved; otherwise we use gateway peering as default, which is the one thing we use it for currently e.g. in NAT tests

Frostman (approved):

PR #1763, opened 2026-05-24
fix(hhfab): wait for server host routes before connectivity probe

pau-hedgehog:
## Summary

Closes the race in `TestConnectivity` where IP discovery (SSH on the server, sees the freshly-leased address immediately) outruns EVPN type-5 propagation across the participating leaves. With a DHCP-churning test running just before, the parallel probe completes inside the convergence gap and fails `sent=N, rcvd=0` on every pair involving the perturbed server.

Adds a gate between IP discovery and the probe loop: for each discovered `/32`, wait until the route appears in the FIB of every leaf attached to its VPC, in the right VRF. 3-minute timeout, log+continue on failure (the gate cannot regress a passing test).

Sits in `TestConnectivity`, so every release-test that calls it (Bundled, MCLAG, Mesh Failover; NAT; multi-VPC) is gated without per-test changes. Reproduced on env-5 (mesh + spine-leaf) and on HLAB CI (`h-gw-iso-l2vni-rt`).

Refs: githedgehog/internal#399

## Test plan

Pre-merge (manual on env-5 + HLAB CI):

- [ ] 5x consecutive release-test single-vpc-suite passes on env-5 / spine-leaf / l3vni
      (each invocation runs DHCP static lease → Bundled Failover, the racing sequence)
- [ ] 5x consecutive release-test single-vpc-suite passes on env-5 / mesh / l3vni
      (covers Bundled and Mesh Failover after the same DHCP static lease preamble)
- [ ] 5x consecutive h-gw-iso-l2vni-rt passes on HLAB CI
      (MCLAG Failover after DHCP static lease)
- [ ] Gate observability: `Waiting for server host routes on leaves` log line appears in runs where the race would previously have fired, with a wait time that resolves before the 3-minute deadline (HLAB serial-SSH polling can take >60s per iteration, hence the wider window); gate is not a no-op
- [ ] No regression in NAT or multi-VPC suites on env-5

Post-merge: `run-vlab-analyzer --failed-only 7` shows zero hits for `Bundled Failover DHCP Race` and `MCLAG Failover DHCP Race` over the following week.

### DIFF
--- a/pkg/hhfab/rt_utils.go
+++ b/pkg/hhfab/rt_utils.go
@@ -9,9 +9,11 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
+	"net/netip"
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	agentapi "go.githedgehog.com/fabric/api/agent/v1beta1"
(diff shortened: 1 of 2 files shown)


pau-hedgehog:
This was a previous attempt to https://github.com/githedgehog/fabricator/pull/1799 but now that we know the underlying root cause it's no use. Closing

Issue #1776, opened 2026-05-27
inspect: LLDP server-uplink neighbor missing on breakout subports (transient, self-recovers)


pau-hedgehog:
## Summary

Fabric `inspect` intermittently fails its LLDP check with a server-uplink neighbor
missing on an E1/1/x breakout subport. L1/L2/L3 are healthy; only the LLDP neighbor
is absent, and it returns on its own shortly after.

Split out from #1760: that issue is the DPB rebind race (neighbor missing
continuously from port recreation, permanent until lldpd restart). This one is the
opposite shape: lost at steady state, hours after any recreation, self-recovering.
Same log line, different mechanism.

## Frequency (VLAB CI Analyzer, githedgehog/lab #364, last 14 days)

9 of the 11 "LLDP Inspect Neighbor Missing" hits are server uplinks (other 2 are
MCLAG, tracked in #1760; gateway uplinks: 0 this window).

```
✗ https://github.com/githedgehog/fabricator/actions/runs/26522099787/job/78116840766 (master: h-gw-iso-l2vni) [env-ci-1.l 05-27 19:31→20:09]  ds3000-03 E1/1/1 server-7
✗ https://github.com/githedgehog/fabricator/actions/runs/26516898957/job/78097605761 (master: h-gw-iso-l2vni) [env-ci-1.l 05-27 18:16→18:54]  ds3000-02 E1/1/2 server-1
✗ https://github.com/githedgehog/fabricator/actions/runs/26507680113/job/78065200293 (master: h-gw-iso-l2vni) [env-ci-1.l 05-27 16:22→17:00]  sse-c4632-01 E1/1/2 server-9
✗ https://github.com/githedgehog/fabric/actions/runs/26397833494/job/77703489853 (master: h-gw-iso-l2vni) [env-ci-1.l 05-25 18:15→18:54]  ds3000-03 E1/1/1 server-7
✗ https://github.com/githedgehog/fabricator/actions/runs/26330415323/job/77515711109 (master: h-gw-iso-l2vni) [env-ci-1.l 05-23 13:12→13:49]  ds3000-03 E1/1/1 server-7
✗ https://github.com/githedgehog/fabricator/actions/runs/26274169979/job/77335044959 (master: h-gw-iso-l2vni-rt) [env-ci-1.l 05-22 10:33→11:11]  ds3000-03 E1/1/1 server-7
[...]

Issue #1760, opened 2026-05-20
SONiC LLDPd silently stops TX on a breakout subport after DPB reconfiguration


pau-hedgehog:
## Summary

Fabric `inspect` fails its LLDP check with a neighbor missing on a fabric/gateway
port that SONiC recreates via Dynamic Port Breakout (DPB) during config apply.
L1/L2/L3 are healthy; only LLDP is absent, and it stays absent through the inspect
window. The server-uplink variant (transient, self-recovering) is a different
mechanism, tracked in #1776.

## Root cause

DPB destroys and recreates the port with a new Linux ifindex. lldpd / lldpmgrd start
before DPB completes and do not re-bind to the recreated interface, so the switch
never receives LLDP on that port even though traffic flows. Failure is time-locked to
the recreation and persists until lldpd restarts.

## Evidence

Original gateway-uplink report, job
https://github.com/githedgehog/fabricator/actions/runs/26123535147/job/76832422566
(in the run log, search `expected neighbor`):

```
21:11:15 ERR Reported  err="switch ds4000-02: E1/20/1: expected neighbor \"gateway-2\" not found"
```

Port recreation with ifindex change is in that run's debug artifact (fab-26123535147-...),
file show-tech-output/ds4000-02-show-tech.log, agent/syslog section (search
`DPB shutdown port Ethernet152` and `key:Ethernet152`): Ethernet152 is deleted and
re-added with a new ifindex while lldpd does not re-bind.

## Frequency (VLAB CI Analyzer, githedgehog/lab #364, last 14 days)

2 of 11 "LLDP Inspect Neighbor Missing" hits are this leg (MCLAG peer link, symmetric).
Unredacted lines in each job log (search `expected neighbor`):

```
✗ https://github.com/githedgehog/fabricator/actions/runs/26277313334/job/77345442174 (ema/test-connectivity: h-gw-iso-l2vni-rt) [env-ci-1.l 05-22 11:11→11:49]
    ERR Reported  err="switch ds3000-01: E1/32/3: expected neighbor \"ds3000-02\" not found"
    ERR Reported  err="switch ds3000-02: E1/32/3: expected neighbor \"ds3000-01\" not found"
✗ https://github.com/githedgehog/fabricator/actions/runs/26147651614/job/76907713011 (master: h-gw-iso-l2vni-rt) [env-ci-1.l 05-20 09:25→10:04]
[...]

PR #1777, opened 2026-05-28
feat(testing): capture LLDP detail on inspect failure (refs #1776)

pau-hedgehog:
## Summary
- Snapshot detailed switch and server LLDP state every time inspect's LLDP retry loop fails an attempt, so transient #1776 failures stop hiding the moment-of-failure state behind a recovered post-inspect show-tech.
- Per-attempt output lives at `show-tech-output/lldp-inspect-captures/attempt-N/<host>.log` so the existing debug bundle picks it up. Best-effort: per-target SSH/exec failures are logged, never propagated.
- Extends `switch.sh` show-tech to always include `show lldp neighbor` plus `lldpcli show neighbors detail / statistics` so non-inspect failure paths get the same evidence.

## Why
Issue #1776 (server-uplink LLDP missing on breakout subports) is transient and self-recovers within seconds. By the time `CollectVLABDebug` runs `show-tech`, the missing neighbor is often back, which is why the existing bundle only contains the summary `show lldp table` and nothing diagnostic. Capturing inside the retry loop preserves a time series: was the neighbor entry never received, present but expiring (TTL countdown), or did the switch only see one side of the LLDP exchange?

## What it captures
Switch side:
- `sonic-cli show lldp neighbor` (full detail with TTL / time-remaining)
- `docker exec lldp lldpcli show neighbors detail`
- `docker exec lldp lldpcli show statistics`
- `docker exec lldp lldpcli show chassis`

Server side (only for affected server-facing legs):
- `networkctl lldp` and `networkctl status`
- `ip -d link show`
- `journalctl -u systemd-networkd -n 300`

## Test plan
This is a diagnostics PR to troubleshoot a recurrent flake on HLAB on CI. After investigation we can assess if anything is worth merging
- [ ] On the next #1776 hit, confirm the artifact bundle contains `lldp-inspect-captures/attempt-{1,2,3}/...` and that it answers: was the neighbor in lldpd's table on the switch? Was the server emitting LLDP frames?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -209,8 +209,9 @@ jobs:
           LOCALREG_SYNC_USERNAME: ${{ github.actor }}
           LOCALREG_SYNC_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          just --timestamp _zot
           just --timestamp _localreg &
-          for i in $(seq 1 30); do curl -sf http://127.0.0.1:30000/v2/ > /dev/null 2>&1 && break || sleep 1; done
+          for i in $(seq 1 60); do curl -sf http://127.0.0.1:30000/v2/ > /dev/null 2>&1 && break || sleep 1; done
           curl -sf http://127.0.0.1:30000/v2/ > /dev/null 2>&1 || { echo "Local registry failed to start"; exit 1; }
           echo "Local registry is ready"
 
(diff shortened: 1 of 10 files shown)


pau-hedgehog:
This PR was instrumental: it captured the moment-of-failure LLDP state that confirmed #1760 is the permanent lldpd-after-DPB symptom on the Celestica Seastone 2 (sse-c4632-01), and that evidence is now staged for the Broadcom case.

The durable, reusable failure-path capture capability has been split into #1782 (native SONiC tech-support collection). The LLDP-specific instrumentation here is not needed on master, so closing. The commits remain on the pau/issue-1776-lldp-instrument branch for audit.

PR #1799, opened 2026-06-10
fix(rt): gate pings on EVPN host-route convergence after DHCP churn

pau-hedgehog:
Fixes githedgehog/internal#399.

After discovering server IPs, wait up to 60 s for each server's /32 host route to appear in the FIB of every attached leaf before issuing pings. Eliminates the race between DHCP-fresh IPs and EVPN type-5 propagation that caused Bundled/Mesh/MCLAG Failover failures.

See the issue for root cause analysis and evidence.

### DIFF
--- a/pkg/hhfab/rt_utils.go
+++ b/pkg/hhfab/rt_utils.go
@@ -305,6 +305,41 @@ func shutDownLinkAndTest(ctx context.Context, testCtx *VPCPeeringTestCtx, link w
 	return DoVLABTestConnectivity(ctx, testCtx.vlabCfg.WorkDir, testCtx.vlabCfg.CacheDir, testCtx.tcOpts)
 }
 
+// getServerVPCAndVRF returns the VPC name and VRF name for the given server by looking up its
+// VPCAttachment. Returns ("", "", nil) if the server has no VPC attachment or the VPC uses L2VNI
+// mode (where host routes are not installed as /32 entries in the IP route table).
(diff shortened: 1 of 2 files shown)


pau-hedgehog:
Closing this. Re-validated internal#399 on the latest Broadcom Enterprise SONiC (4.5.2, build `3cd846de9f`) on env-1: the orchagent `handleConflictingNeighbor` conflict is not fixed, and the data-plane impact is worse than first seen, a ~5 minute black-hole that recovers only when the Drop neighbor ages out (no fast retry). Full re-validation and evidence in internal#399.

This convergence gate does not hold as a fix:

- It is a no-op in L2VNI: `getServerVPCAndVRF` returns empty for `VPCModeL2VNI`, so `waitForServerHostRoutesInLeaves` skips entirely. The green CI signal comes only from the mode where the gate never runs.
- In L3VNI it does not mask the bug, it relocates the failure. On DS5000 (env-5) the /32 stays out of the RIB and the 60s gate times out (seen on the "No restrictions" test: route `10.50.2.4/32` on `ds5000-01`, `context deadline exceeded` at exactly 60s). On S5248 (env-1) the /32 is in the RIB but the ASIC is mis-programmed, so the gate passes in ~6s and pings fail.
- 60s is shorter than the real recovery (~5 min), and `show ip route` (RIB) does not reflect the ASIC state where this bug actually lives.

The fix belongs in SONiC. Filing a Broadcom case; tracking under internal#399. We can revisit a tracked known-issue accommodation if we need to unblock release CI before the SONiC fix lands.

PR #1808, opened 2026-06-12
test(release): ESLAG fallback review nits

pau-hedgehog:
Follow-up to #1759 addressing edipascale's review comments.

- rename `findESLAGAttachedSubnet` -> `findESLAGAttachedVLAN` (name reflects what it returns)
- filter `VPCAttachment` list by connection label instead of post-filtering in the loop
- drop `eslagPxeAttemptsPerLegX`; no reason for a higher attempt count in extended mode
- rename `ranAtLeastOne` -> `ranAtLeastOnce`; move assignment inside the slave loop so an empty `slaves` slice doesn't count as "ran"
- fix misleading `timeout(1)` reference in `pxeAttempt` comment

### DIFF
--- a/pkg/hhfab/rt_eslag_fallback.go
+++ b/pkg/hhfab/rt_eslag_fallback.go
@@ -18,9 +18,8 @@ import (
 )
 
 const (
-	eslagPxeDHCPTimeoutSec  = 20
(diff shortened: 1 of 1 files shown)


edipascale (approved):
thanks!

PR #1813, opened 2026-06-15
bump(nos): bcm to 4.5.2

Frostman:


pau-hedgehog:
Release tests pass both in env-1 spine-leaf and mesh, @Frostman

PR #1819, opened 2026-06-16
fix(show-tech): capture flatcar-install service status in gateway console fallback

pau-hedgehog:
Adds `systemctl status flatcar-install.service` and `journalctl -u flatcar-install` to the gateway console fallback probe.

The service is only present during USB install (no-op on installed systems). Without it, a gateway install hang is undiagnosable from the console fallback output — only the disk partition kernel messages and an idle load average are visible, with no indication of where the installer stopped.

Found during triage of run 27566110480 (`v-gw-onie-usb-l2vni-rt`): gateway-2 wrote disk at 168s then went idle for 40 min without rebooting; root cause was indeterminate because the installer service logs were not collected.


pau-hedgehog:
superseded: pushing directly to pau/extend-show-tech (PR #1630)

PR #1836, opened 2026-06-23
support unbundled multihomed servers in SetupVPCs / TestConnectivity

edipascale:
stacked on top of the connectivity matrix refactor

- in SetupVPCs, automatically opt for a HostBGP subnet if the server has multiple unbundled connections
- handle this case in TestConnectivity, i.e. when discovering the endpoints
- support multihomed servers on a mix of eslag leaves and orphan leaves if there are not enough orphan leaves, e.g. in our standard topology

### DIFF
--- a/cmd/hhfab/main.go
+++ b/cmd/hhfab/main.go
@@ -1431,7 +1431,7 @@ Examples:
 								return err
 							}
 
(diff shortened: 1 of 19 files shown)


edipascale:
closing in favor of https://github.com/githedgehog/fabricator/pull/1910

PR #1843, opened 2026-06-29
Harden SSH config for control node

mrbojangles3:
related https://github.com/githedgehog/internal/issues/370

### DIFF
--- a/api/fabricator/v1beta1/fabricator_types.go
+++ b/api/fabricator/v1beta1/fabricator_types.go
@@ -221,6 +221,9 @@ type ControlConfig struct {
 	DummySubnet meta.Prefix `json:"dummySubnet,omitempty"`
 
 	DefaultUser ControlUser `json:"defaultUser,omitempty"`
+	// NoPassAuth disables SSH password authentication on control/fab nodes, requiring key-based access.
+	// When true, at least one authorized key must be configured.
+	NoPassAuth bool `json:"noPassAuth,omitempty"`
 
 	NTPServers []string `json:"ntpServers,omitempty"`
 
@@ -818,6 +821,9 @@ func (f *Fabricator) Validate(ctx context.Context) error {
 		ph := f.Spec.Config.Control.DefaultUser.PasswordHash
 		if ph == "" {
 			return fmt.Errorf("default control user password hash is required") //nolint:goerr113
+		}
+		if f.Spec.Config.Control.NoPassAuth && len(f.Spec.Config.Control.DefaultUser.AuthorizedKeys) == 0 {
+			return fmt.Errorf("noPassAuth is true; but no default authorized keys are configured") //nolint:goerr113
 		}
 
 		if !strings.HasPrefix(ph, "$5$") {
--- a/pkg/fab/initconfig.tmpl.yaml
+++ b/pkg/fab/initconfig.tmpl.yaml
@@ -39,6 +39,8 @@ spec:
       #joinToken: super-secret-join-token
       {{- end }}
 
+      noPassAuth: {{ .NoPassAuth }} # disable SSH password authentication (requires authorizedKeys below)
+
       defaultUser: # user 'core' on all control nodes
         {{- if .DefaultPasswordHash }}
         password: "{{ .DefaultPasswordHash }}" # password hash
@@ -123,7 +125,7 @@ spec:
    disk: "/dev/sda" # disk to install OS on, e.g. "sda" or "nvme0n1"
   external:
     interface: enp2s0 # interface for external
-    ip:	dhcp # IP address for external interface
+    ip: dhcp # IP address for external interface
   management:
     interface: enp2s1 # interface for management
 
--- a/pkg/fab/recipe/control_butane.tmpl.yaml
+++ b/pkg/fab/recipe/control_butane.tmpl.yaml
@@ -124,6 +124,11 @@ storage:
           Destination=0.0.0.0/0
           Metric=42000
 
+    - path: /etc/ssh/sshd_config.d/40-hedgehog.conf
+      mode: 0600
+      contents:
+        inline: |{{ .SSHDConfig | nindent 10 }}
+
     - path: /etc/systemd/network/20-mgmt.network
       mode: 0644
       contents:
--- a/pkg/fab/recipe/control_upgrade.go
+++ b/pkg/fab/recipe/control_upgrade.go
@@ -168,6 +168,9 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
 	if err := c.installFabricCtl(); err != nil {
 		return fmt.Errorf("installing kubectl-fabric: %w", err)
 	}
+	if err := c.installSSHConfig(); err != nil {
+		return fmt.Errorf("installing ssh configuration: %w", err)
+	}
 
 	if err := copyFile(k9s.BinName, filepath.Join(k3s.BinDir, k9s.BinName), 0o755); err != nil {
 		return fmt.Errorf("copying k9s bin: %w", err)
@@ -540,6 +543,26 @@ func (c *ControlUpgrade) setupFirewall(ctx context.Context) error {
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("error on systemctl enable --now %s: %w", nftServiceName, err)
 	}
+
+	return nil
+}
+
+func (c *ControlUpgrade) installSSHConfig() error {
+	// We do not support making changes and doing upgrades in the same step
+	// So this configuration comes from the running instance.
+	// Users will need to manually place the option or wait for the upgrade to 26.04
+	fileContents, err := renderSSHDConfig(c.Fab.Spec.Config.Control.NoPassAuth)
+	if err != nil {
+		return fmt.Errorf("rendering sshd config: %w", err)
+	}
+
+	if err := os.WriteFile(sshdConfigPath, []byte(fileContents+"\n"), 0o600); err != nil {
+		return fmt.Errorf("writing ssh config: %w", err)
+	}
+	// There is no need to reload the sshd. flatcar linux uses socket activation for sshd
+	// which is similar to inetd mode. Each new ssh connection reads the configuration files on start
+	// So the next connection will be using the new config after it is written.
+	// Restarting existing connections is undesirable in an upgrade context.
 
 	return nil
 }
--- a/pkg/fab/recipe/sshd.go
+++ b/pkg/fab/recipe/sshd.go
@@ -1,0 +1,51 @@
+// Copyright 2024 Hedgehog
+// SPDX-License-Identifier: Apache-2.0
+
+package recipe
+
+import (
+	"fmt"
+
+	"go.githedgehog.com/fabricator/pkg/util/tmplutil"
+)
+
+// sshdConfigPath is the sshd drop-in written on both control and fab nodes.
+const sshdConfigPath = "/etc/ssh/sshd_config.d/40-hedgehog.conf"
+
+// sshdConfigTmpl is the single source of truth for the contents of 40-hedgehog.conf.
+// It is rendered for both fresh installs (injected into the butane templates) and
+// upgrades (written to disk directly), so the two paths can never drift apart.
+// Keep the closing backtick on the MACs line so the body has no trailing newline.
+const sshdConfigTmpl = `# Managed by hh, do not change
+PerSourceMaxStartups 4
+PerSourcePenalties crash:5m authfail:2m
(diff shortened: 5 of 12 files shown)


pkg/fab/recipe/control_butane.tmpl.yaml
```diff
@@ -112,6 +112,123 @@ storage:
...
+          #IgnoreRhosts yes
+
+          # To disable tunneled clear text passwords, change to "no" here!
+          PasswordAuthentication no
+          #PermitEmptyPasswords no
+
+          # Change to "no" to disable keyboard-interactive authentication.  Depending on
+          # the system's configuration, this may involve passwords, challenge-response,
+          # one-time passwords or some combination of these and other methods.
+          KbdInteractiveAuthentication no
```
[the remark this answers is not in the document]
mrbojangles3: This is a good idea. We should force the installers to use SSH.

Frostman (changes requested):
It's not going to be applied during the upgrade, and I'm not sure we enforce having SSH keys configured. I believe some of our customers are using password auth to access control node for some reason.

I suggest two things:
1. make it applied on upgrade w/o restricting password auth
2. add flag to the fabricator config to prohibit password auth that is set by default and requires non-empty ssh keys. it'll mean that for all new deployments we'll try to enforce it; and when flag is set we can enforce no password auth in ssh config

pkg/fab/recipe/control_butane.tmpl.yaml
```diff
@@ -112,6 +112,123 @@ storage:
           Gateway={{ .DummyGateway }}
           Destination=0.0.0.0/0
           Metric=42000
+    - path: /etc/ssh/sshd_config
```
Frostman: I think it'll be better to put it as a drop-in and only include meaningful configs with useful commens, not just copy-paste of the full ssh cconfig example

Frostman:
1. `hhfab init --ctrl-no-pass-auth=true` (it's true by default, so people can disable it). It should be wired through initconfig.go
1. if `--dev` is enabled, password auth should be enabled (just "reset" logic in the initconfig.go)
1. if pass auth disabled, ssh keys must not be empty
1. it should result in the Fabricator config; it's already in the right location, but should be inverted like `ControlConfig.NoPassAuth`

pkg/fab/recipe/control_butane.tmpl.yaml
```diff
@@ -113,6 +113,27 @@ storage:
           Destination=0.0.0.0/0
           Metric=42000
 
+    - path: /etc/ssh/sshd_config.d/40-hedgehog.conf
```
Frostman, line 127: Add go text tmpl comment pointing to the other place same config is used or move to const.

```
{{/* This is a comment in a Go template */}}
```

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,32 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
 	return nil
 }
 
+func copySSHConfig(_ context.Context) error {
```
Frostman: Add go comment pointing to the other place same config is used or move to const.

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,32 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
...
+	fileContents := `
+# Customized by Hedgehog
+PerSourceMaxStartups 4
+PerSourcePenalties crash:5m authfail:2m
+LoginGraceTime 20
+PermitRootLogin no
+MaxAuthTries 3
+MaxSessions 5
+PasswordAuthentication yes
```
Frostman: This should come from fabrictor config

pkg/fab/recipe/node_build.go
```diff
@@ -122,6 +122,7 @@ func (b *NodeInstallBuilder) addPayload(ctx context.Context, slog *slog.Logger,
 var nodeButaneTmpl string
 
 func (b *NodeInstallBuilder) buildIgnition() ([]byte, error) {
+	const sshdYes = "yes"
```
Frostman: doesn't make sense to make a const in a func for it

Frostman (changes requested):

pkg/fab/recipe/control_upgrade.go
```diff
@@ -173,6 +174,19 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
...
+	// (the pre-upgrade cluster CR). When upgrading from a version without this field, the
+	// cluster CR can't carry it across the version boundary, so the operator's intent for
+	// the target OS lives only in the bundled config.
+	noPassAuth, err := bundledNoPassAuth(ctx, c.WorkDir)
+	if err != nil {
+		return fmt.Errorf("reading bundled ssh config: %w", err)
+	}
+
+	if err := copySSHConfig(ctx, noPassAuth); err != nil {
+		return fmt.Errorf("copying ssh config:%w", err)
```
mrbojangles3: add a space here

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,22 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
...
+	if err != nil {
+		return fmt.Errorf("rendering ssh config: %w", err)
+	}
+
+	if err := os.WriteFile(sshdConfigPath, []byte(fileContents+"\n"), 0o600); err != nil {
+		return fmt.Errorf("writing ssh config: %w", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
mrbojangles3: This should happen and its not.

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,22 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
...
+	// password auth enabled so an upgrade never locks an operator out. When the value is
+	// present it is honored, disabling password auth on the upgraded node.
+	fileContents, err := renderSSHDConfig(noPassAuth)
+	if err != nil {
+		return fmt.Errorf("rendering ssh config: %w", err)
+	}
+
+	if err := os.WriteFile(sshdConfigPath, []byte(fileContents+"\n"), 0o600); err != nil {
+		return fmt.Errorf("writing ssh config: %w", err)
+	}
```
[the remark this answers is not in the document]
mrbojangles3: flatcar is the only supported OS

api/fabricator/v1beta1/fabricator_types.go
```diff
@@ -819,6 +822,10 @@ func (f *Fabricator) Validate(ctx context.Context) error {
 		if ph == "" {
 			return fmt.Errorf("default control user password hash is required") //nolint:goerr113
 		}
+		sshKey := f.Spec.Config.Control.DefaultUser.AuthorizedKeys
+		if len(sshKey) == 0 {
+			return fmt.Errorf("must have default ssh key") //nolint:goerr113
+		}
```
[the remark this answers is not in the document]
Frostman: @mrbojangles3 that's correct and needs to be fixed

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,58 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
 	return nil
 }
 
+func copySSHConfig(ctx context.Context, noPassAuth bool) error {
```
Frostman: You should just make it a method of `ControlUpgrade` and use `c.Fab`, as it doesn't really matter what embedded and live state from the cluster would be the same default `false` for the NoPassAuth. It means `bundledNoPassAuth` could be removed altogether.

It's probably better to name it `installSSHConfig` as it's not just copying the file but reloading the daemon too.

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,58 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
...
+		// sockets don't support reload, they get restarted instead
+		cmd := exec.CommandContext(ctx, "systemctl", "reload-or-restart", unit)
+		cmd.Stdout = logutil.NewSink(ctx, slog.Debug, "systemctl: ")
+		cmd.Stderr = logutil.NewSink(ctx, slog.Debug, "systemctl: ")
+
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("reloading %q: %w", unit, err)
+		}
+
+		return nil
```
Frostman: Are you sure you want to return from reloadSSHD on a first success? Means you only reload first unit always or fail with error on it

pkg/fab/recipe/install_upgrade.go
```diff
@@ -247,6 +247,58 @@ func DoUpgrade(ctx context.Context, workDir string, yes, skipChecks bool) error
...
+	}
+
+	if err := reloadSSHD(ctx); err != nil {
+		return fmt.Errorf("reloading sshd: %w", err)
+	}
+
+	return nil
+}
+
+var sshdUnits = []string{"sshd.socket", "sshd.service"}
```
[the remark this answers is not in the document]
Frostman: Do we even need to reload socket?

pkg/fab/initconfig.tmpl.yaml
```diff
@@ -39,6 +39,10 @@ spec:
       #joinToken: super-secret-join-token
       {{- end }}
 
+      {{- if .NoPassAuth }}
```
Frostman: I think it'll be nice to always generate it on init, even if it's false, and keep the comment so users know what config to touch.

pkg/fab/initconfig.tmpl.yaml
```diff
@@ -123,7 +127,7 @@ spec:
    disk: "/dev/sda" # disk to install OS on, e.g. "sda" or "nvme0n1"
   external:
     interface: enp2s0 # interface for external
-    ip:	dhcp # IP address for external interface
+    ip: dhcp # IP address for external interface
```
Frostman, line 128: Scary - what changed here?

pkg/fab/recipe/sshd.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// sshdConfigPath is the sshd drop-in written on both control and fab nodes.
+const sshdConfigPath = "/etc/ssh/sshd_config.d/40-hedgehog.conf"
+
+// sshdConfigTmpl is the single source of truth for the contents of 40-hedgehog.conf.
+// It is rendered for both fresh installs (injected into the butane templates) and
+// upgrades (written to disk directly), so the two paths can never drift apart.
+// Keep the closing backtick on the MACs line so the body has no trailing newline.
+const sshdConfigTmpl = `# Customized by Hedgehog
```
Frostman: Let's make it smth like "Managed by hh, do not change"

Frostman (changes requested):

pkg/fab/initconfig.tmpl.yaml
mrbojangles3: My vim editor shows a non printable character here '>' but can try a hex dump to see whats in here. Could be a problem in my setup.

api/fabricator/v1beta1/fabricator_types.go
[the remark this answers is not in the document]
mrbojangles3: I will change the error message, and add check.

pkg/fab/recipe/install_upgrade.go
mrbojangles3: - [ ] Change name
- [ ] Dont read setting from passed in file, only use the runtime config as we don't want to make a user change at the same time as an upgrade.

pkg/fab/initconfig.tmpl.yaml
mrbojangles3: AI mentioned it was a tab not a space

pkg/fab/recipe/node_butane.tmpl.yaml
```diff
@@ -74,6 +74,11 @@ storage:
         inline: |
           Hedgehog Node managed by Fabricator
 
+    - path: /etc/ssh/sshd_config.d/40-hedgehog.conf
```
Frostman: We should either not do it for nodes at all or add handling during node upgrade. Doing only on the node install will lead to inconsistent state in production after upgrades.

Frostman (changes requested):

Frostman (approved):

PR #1891, opened 2026-07-24
close ports on control node, keep ssh and kubeapi open

mrbojangles3:
Fixes githedgehog/internal#371

### DIFF
--- a/pkg/fab/comp/k3s/server_config.tmpl.yaml
+++ b/pkg/fab/comp/k3s/server_config.tmpl.yaml
@@ -4,6 +4,8 @@ flannel-iface: "{{ .FlannelIface }}"
 cluster-cidr: "{{ .ClusterSubnet }}"
 service-cidr: "{{ .ServiceSubnet }}"
 cluster-dns: "{{ .ClusterDNS }}"
+kube-proxy-arg:
+  - "nodeport-addresses={{ .NodeSubnet }}"
 write-kubeconfig-mode: "0644"
 tls-san:
   {{ range .TLSSAN }}
--- a/pkg/fab/recipe/control_upgrade.go
+++ b/pkg/fab/recipe/control_upgrade.go
@@ -173,6 +173,10 @@ func (c *ControlUpgrade) Run(ctx context.Context) error {
 		return fmt.Errorf("copying k9s bin: %w", err)
 	}
 
+	if err := c.setupFirewall(ctx); err != nil {
+		return fmt.Errorf("setup firewall: %w", err)
+	}
+
 	if err := upgradeFlatcar(ctx, string(flatcar.Version(c.Fab)), c.Yes); err != nil {
 		return fmt.Errorf("upgrading Flatcar: %w", err)
 	}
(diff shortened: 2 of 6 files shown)


pkg/fab/recipe/control_upgrade.go
```diff
@@ -497,6 +501,42 @@ func (c *ControlUpgrade) installFabricCtl() error {
...
+	}
+
+	cmd := exec.CommandContext(ctx, "systemctl", "daemon-reload")
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error on systemctl daemon-reload: %w", err)
+	}
+	cmd = exec.CommandContext(ctx, "systemctl", "enable", "--now", nftServiceName)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error on systemctl enable --now %s: %w", nftServiceName, err)
+	}
```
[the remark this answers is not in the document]
mrbojangles3: this is a oneshot file. an enable now, will send the rules in.

mrbojangles3:
- [ x ]: Test an update on a system that already has the rules so we can see if there is a problem with the loading mechanism

pkg/fab/recipe/nftables.go
```diff
@@ -0,0 +1,81 @@
...
+const (
+	nftablesRulesFilePath = "/etc/hh/nftables.conf"
+	nftUnitFilePath       = "/etc/systemd/system/hh-nftables.service"
+	nftServiceName        = "hh-nftables.service"
+)
+
+const nftablesTmpl = `#HH table to secure the external port
+nft flush table inet HH-ext-intf-fw
+table inet HH-ext-intf-fw {
+    counter ext_dropped {
```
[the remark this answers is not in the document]
mrbojangles3: This is a good comment. I have changed the verb to `destroy` per the nft man page. there are no errors if the table doesn't exist.

pkg/fab/recipe/nftables.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+
+	"go.githedgehog.com/fabricator/pkg/util/tmplutil"
+)
+
+// nftablesRulesFilePath and nftUnitFilePath are the two files written on both
+// control/fab nodes. They are the single source of truth for these paths: the
+// butane templates (fresh install) hardcode the same strings and the upgrade
+// path writes them here, so the install and upgrade paths can never drift apart.
+// nftServiceName must match the unit name declared in the butane templates.
```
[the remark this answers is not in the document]
mrbojangles3: done

pkg/fab/recipe/nftables.go
```diff
@@ -0,0 +1,80 @@
...
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ProtectSystem=full
+ProtectHome=true
+ExecStart=/usr/bin/nft -f {{ .RulesPath }}
+ExecReload=/usr/bin/nft -f {{ .RulesPath }}
+ExecStop=/usr/bin/nft delete table inet HH-ext-intf-fw
```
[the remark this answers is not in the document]
mrbojangles3: Delete is no longer used. Destroy is

pkg/fab/comp/k3s/server_config.tmpl.yaml
```diff
@@ -4,6 +4,8 @@ flannel-iface: "{{ .FlannelIface }}"
 cluster-cidr: "{{ .ClusterSubnet }}"
 service-cidr: "{{ .ServiceSubnet }}"
 cluster-dns: "{{ .ClusterDNS }}"
+kube-proxy-arg:
+  - "nodeport-addresses={{ .NodeSubnet }}"
```
Frostman, line 8: It feels like we'll have to change it in the future when we need to expose the service on the `ext` interface, but that's not a big deal to change later

pkg/fab/recipe/nftables.go
```diff
@@ -0,0 +1,81 @@
...
+// butane templates (fresh install) hardcode the same strings and the upgrade
+// path writes them here, so the install and upgrade paths can never drift apart.
+// nftServiceName must match the unit name declared in the butane templates.
+const (
+	nftablesRulesFilePath = "/etc/hh/nftables.conf"
+	nftUnitFilePath       = "/etc/systemd/system/hh-nftables.service"
+	nftServiceName        = "hh-nftables.service"
+)
+
+const nftablesTmpl = `#HH table to secure the external port
```
Frostman: smth like "managed by hedgehog, do not edit" similar to ssh dropin would be nice here

pkg/fab/recipe/control_upgrade.go
```diff
@@ -497,6 +501,42 @@ func (c *ControlUpgrade) installFabricCtl() error {
...
+
+	nftUnitFile, err := renderNftablesSystemdUnit()
+	if err != nil {
+		return fmt.Errorf("rendering nftables service unit: %w", err)
+	}
+	if err := os.WriteFile(nftUnitFilePath, []byte(nftUnitFile), 0o600); err != nil {
+		return fmt.Errorf("writing nftables service unit: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, "systemctl", "daemon-reload")
```
Frostman, line 530: It's better to add

```
	cmd.Stdout = logutil.NewSink(ctx, slog.Debug, "systemctl: ")
	cmd.Stderr = logutil.NewSink(ctx, slog.Debug, "systemctl: ")
```

for both commands, b/c otherwise we'll lose all logs from it and it'll be impossible to debug.

Don't forget to start function with dedicating nested context with timeout so we don't stuck on it and sinks would be cleaned.

```
	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
	defer cancel()
```

Frostman (changes requested):

pkg/fab/recipe/control_upgrade.go
mrbojangles3: done

pkg/fab/recipe/nftables.go
mrbojangles3: done

Frostman (approved):

PR #1905, opened 2026-07-29
fix: add firewall rules for gw mgmt intf

### LINKED ISSUE #1904: Only expose SSH and K8s API on gatewaynodes
Similar to https://github.com/githedgehog/fabricator/pull/1891

mrbojangles3:
closes #1904

### DIFF
--- a/pkg/fab/recipe/node_build.go
+++ b/pkg/fab/recipe/node_build.go
@@ -134,12 +134,22 @@ func (b *NodeInstallBuilder) buildIgnition() ([]byte, error) {
 	if dummyIP.Bits() != 31 {
 		return nil, fmt.Errorf("dummy IP must be a /31") //nolint:goerr113
 	}
+	nftRules, err := renderNftablesRules(b.Node.Spec.Management.Interface)
+	if err != nil {
+		return nil, fmt.Errorf("rendering nftables rules: %w", err)
+	}
+	nftUnitFile, err := renderNftablesSystemdUnit()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/fab/recipe/node_install.go
```diff
@@ -237,6 +240,49 @@ EmitLLDP=yes
 	return nil
 }
 
+func (c *NodeInstallUpgrade) installNftRules(ctx context.Context) error {
```
Frostman, line 243: Let's dedup the code with the control branch. It sounds like converting ControlUpgrade.setupFirewall to a function accepting all we need and calling it from both places

Frostman (changes requested):

pkg/fab/recipe/node_build.go
```diff
@@ -134,12 +134,22 @@ func (b *NodeInstallBuilder) buildIgnition() ([]byte, error) {
 	if dummyIP.Bits() != 31 {
 		return nil, fmt.Errorf("dummy IP must be a /31") //nolint:goerr113
 	}
+	nftRules, err := renderNftablesRules(b.Node.Spec.Management.Interface)
```
Frostman, line 137: We're only doing it for the external interface on the control node, I think we don't need it now on the gateways/other nodes as we don't have an external interfaces there (yet)

Frostman (changes requested):

Frostman:
We aren't doing anything for the mgmt port on control node(s) as of now, so no need to do it on the nodes/gateways

PR #1910, opened 2026-07-29
test-connectivity follow-up + multihomed unbundled + ACLs

### LINKED ISSUE #1873: Add release-tests for new gateway ACLs

edipascale:
An attempt to address most of the comments on https://github.com/githedgehog/fabricator/pull/1761, kept as a separate branch in case we decide to merge the connectivity test refactor as it currently is while we address them.

Includes a manual pass to clean a big chunk of the AI comment slop. I kept what I thought was useful (e.g. gotchas and clarifications).

Also includes other stuff that was stacked on top of test-connectivity as separate PRs (i.e. multihomed unbundled support and ACLs release tests), since it was asked by the reviewers.

Does not include the hostbgp release test for the moment, as rebasing it on top of the connectivity refactor implies substantial changes and I want to make sure I'm not breaking it before committing to the change

### DIFF
--- a/pkg/hhfab/endpoints.go
+++ b/pkg/hhfab/endpoints.go
@@ -20,14 +20,8 @@ import (
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// SSHResolver returns the SSH config for a server. The collector and per-
-// server refresh helpers take one of these instead of a pre-built
-// map[string]*sshutil.Config so callers can share whatever SSH plumbing
-// they already have (SetupVPCs builds a map keyed by VM name; rt_* tests
-// resolve lazily through testCtx.getSSH).
 type SSHResolver func(server string) (*sshutil.Config, error)
 
-// SSHResolverFromMap adapts a pre-built map into an SSHResolver.
 func SSHResolverFromMap(m map[string]*sshutil.Config) SSHResolver {
 	return func(server string) (*sshutil.Config, error) {
 		cfg, ok := m[server]
@@ -39,9 +33,6 @@ func SSHResolverFromMap(m map[string]*sshutil.Config) SSHResolver {
 	}
 }
 
-// discoveredIP pairs a server-side interface name with the address found
-// on it. The interface tells us whether the address is a hostBGP /32 VIP
-// (on `lo`) or a regular subnet address (on the bond/VLAN interface).
 type discoveredIP struct {
 	iface  string
 	prefix netip.Prefix
@@ -78,12 +69,8 @@ func discoverServerIPs(ctx context.Context, sshCfg *sshutil.Config, server strin
 }
 
 // DroppedEndpoint records an attachment or address that endpoint discovery
-// could not turn into a testable *Endpoint. A drop is not an error on its
-// own — plain `hhfab vlab setup-vpcs` tolerates it — but it does mean some
-// part of the topology is invisible to the connectivity matrix, and
-// ConnectivityMatrix.Validate refuses to run a test against a matrix that
-// carries any. VPC/Subnet are empty when the drop is about an address that
-// matched no attachment.
+// could not turn into a testable *Endpoint. ConnectivityMatrix.Validate()
+// refuses to run a test against a matrix that carries any of these.
 type DroppedEndpoint struct {
 	Server string
 	VPC    string
@@ -99,8 +86,6 @@ func (d DroppedEndpoint) String() string {
 	return fmt.Sprintf("%s (%s/%s): %s", d.Server, d.VPC, d.Subnet, d.Reason)
 }
 
-// serverAttachment is the (vpc, subnet) information the collector resolves
-// from a VPCAttachment + its referenced VPC CRD.
 type serverAttachment struct {
 	vpcName    string
 	subnetName string
@@ -112,25 +97,6 @@ type serverAttachment struct {
 // CollectServerEndpoints observes the live cluster and produces one
 // *Endpoint per (server, vpc, subnet) attachment for each server in the
 // `servers` filter (nil → all servers attached to at least one VPC).
-//
-// Algorithm: list VPCAttachments → group by server (resolved via each
-// attachment's Connection); for each candidate server, SSH it and read all
-// IPv4 addresses; match each address to one of its attachments by CIDR
-// containment (narrowest prefix wins on ties). HostBGP is derived from
-// VPCSubnet.HostBGP.
-//
-// The second return value lists everything discovery could not represent as
-// an endpoint. Two conditions are reported there: an attachment with no
-// matching address on a server that did report addresses, and an address
-// matching none of its server's attachments. Both mean something in the
-// topology would go untested, so ConnectivityMatrix.Validate rejects them.
-// A server with no configured addresses at all is only warned about and is
-// not reported as a drop — this matches today's SetupVPCs behavior, where
-// ESLAG servers in L3VNI mode are skipped before hhnet runs and therefore
-// legitimately have no IPs.
-//
-// Returns an error only when SSH itself fails on a queried server or when a
-// VPCAttachment cannot be resolved to a Connection.
 func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHResolver, servers []string) ([]*Endpoint, []DroppedEndpoint, error) {
 	attaches := &vpcapi.VPCAttachmentList{}
 	if err := kube.List(ctx, attaches); err != nil {
@@ -181,10 +147,6 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 			return nil, nil, fmt.Errorf("getting endpoints of connection %q: %w", conn.Name, err)
 		}
 		if len(srvs) != 1 {
-			// VPCAttachments only reference server-facing connections; if a
-			// connection has no server endpoint we skip it as a malformed
-			// attachment rather than erroring out, since fabric webhooks
-			// already gate this.
 			continue
 		}
 		serverName := srvs[0]
@@ -215,8 +177,6 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 		})
 	}
 
-	// Probe every candidate server in parallel; errgroup mirrors what
-	// SetupVPCs does for the hhnet config loop.
 	type collected struct {
 		serverName string
 		ips        []discoveredIP
@@ -253,10 +213,7 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 	}
 
 	// Build endpoints by matching each discovered IP against the server's
-	// candidate attachments (already narrowed by VPCAttachment list). When
-	// multiple attachments contain the IP, the narrowest-prefix attachment
-	// wins (handles the P2P /31 case where a /31 sits inside the parent
-	// /24).
+	// candidate attachments; narrowest-prefix attachment wins
 	slices.SortFunc(probed, func(a, b collected) int { return strings.Compare(a.serverName, b.serverName) })
 	out := []*Endpoint{}
 	dropped := []DroppedEndpoint{}
@@ -264,8 +221,7 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 		atts := serverAttachments[p.serverName]
 		used := make([]bool, len(atts))
 		if len(p.ips) == 0 {
-			// Expected for ESLAG servers in L3VNI mode, which never run
-			// hhnet; not recorded as a drop (see the function doc).
+			// Expected for ESLAG servers in L3VNI mode, which never run hhnet; not recorded as a drop
 			slog.Warn("Server has no configured IPs, skipping endpoints", "server", p.serverName, "attachments", len(atts))
 
 			continue
@@ -292,8 +248,6 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 				continue
 			}
 			if used[bestIdx] {
-				// The attachment still has an endpoint (the first matching
-				// address won), so nothing goes untested — warn only.
 				slog.Warn("Multiple server IPs match the same attachment, keeping the first one",
 					"server", p.serverName, "iface", ip.iface, "addr", ip.prefix.String(),
 					"vpc", atts[bestIdx].vpcName, "subnet", atts[bestIdx].subnetName)
@@ -332,26 +286,10 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 
 // ReplaceServerEndpoints reconciles the matrix's endpoints for one
 // server against newEPs:
-//
 //   - Existing endpoints whose (vpc, subnet) match an entry in newEPs
-//     are updated in place — the IP and HostBGP fields are copied
-//     across, but the *Endpoint pointer stays the same. Matrix entries
-//     keyed on that pointer remain valid, so suite setups that don't
-//     wipe between tests keep their verdicts.
-//   - Existing endpoints with no (vpc, subnet) match in newEPs are
-//     dropped from AllEndpoints; entries referencing them are deleted
-//     (the topology those verdicts assumed no longer holds).
-//   - newEPs that don't match any existing endpoint are appended; the
-//     matrix has no entries for them until the next Repopulate or
-//     overlay.
-//
-// Use after a runtime topology change (server moved to a different VPC,
-// or DHCP lease refreshed) to bring the matrix back in sync with the
-// cluster while preserving entries for attachments that didn't change.
-//
-// Entries in newEPs whose Server.Name != name are appended as-is, so
-// mixing External endpoints into the slice is a no-op for the matching
-// pass.
+//     are updated in place
+//   - Existing endpoints with no (vpc, subnet) match in newEPs are dropped
+//   - newEPs that don't match any existing endpoint are appended
 func (m *ConnectivityMatrix) ReplaceServerEndpoints(name string, newEPs []*Endpoint) {
 	if m == nil {
 		return
@@ -409,10 +347,6 @@ func (m *ConnectivityMatrix) ReplaceServerEndpoints(name string, newEPs []*Endpo
 	}
 }
 
-// ReplaceServerDrops swaps the matrix's recorded discovery drops for one
-// server with the ones from a fresh CollectServerEndpoints sweep. Pair it
-// with ReplaceServerEndpoints so a rebind can't leave the matrix carrying
-// drops from a topology the server no longer has (or hide new ones).
 func (m *ConnectivityMatrix) ReplaceServerDrops(name string, dropped []DroppedEndpoint) {
 	if m == nil {
 		return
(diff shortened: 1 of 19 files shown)


pkg/hhfab/rt_acl_tests.go
```diff
@@ -0,0 +1,489 @@
...
+		{Name: "Gateway Peering ACL Explicit Allow", F: gatewayACLExplicitAllowTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Protocol Scoping", F: gatewayACLProtocolScopingTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Packet One-Way", F: gatewayACLPacketOneWayTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Flow Scope Masquerade", F: gatewayACLFlowScopeMasqueradeTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Subnet/CIDR Scoping", F: gatewayACLSubnetScopingTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Port Range Scoping", F: gatewayACLPortScopingTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Precedence Allow-Then-Deny", F: gatewayACLPrecedenceAllowThenDenyTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+		{Name: "Gateway Peering ACL Precedence Deny-Then-Allow", F: gatewayACLPrecedenceDenyThenAllowTest, SkipFlags: SkipFlags{NoGateway: true, NoServers: true}},
+	}
+}
```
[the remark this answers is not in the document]
edipascale: the comment is outdated, dataplane has been bumped and the tests are expected to pass now - I'll modify the comments and remove some of the excessive AI comments altogether

pkg/hhfab/rt_single_vpc_suite.go
```diff
@@ -456,6 +456,16 @@ func gatewayFailoverTest(ctx context.Context, testCtx *VPCPeeringTestCtx, matrix
...
 
+	// The suite-level matrix predates this peering; refresh it so the
+	// connectivity checks below (during failover, and in the reverts while
+	// the peering is still up) assert the peered state. Reverts run in
+	// reverse order, so the teardown registered above runs after them.
+	// Repopulating only once the teardown is registered keeps a failure
+	// here from leaking the peering into the next test.
+	if err := matrix.Repopulate(ctx, testCtx.kube); err != nil {
+		return false, reverts, fmt.Errorf("refreshing matrix after gateway peering: %w", err)
+	}
```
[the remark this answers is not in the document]
edipascale: every other test in the suite calls repopulate after setting its own peerings, so this has no practical consequences

pkg/hhfab/matrix.go
```diff
@@ -332,6 +366,110 @@ func (m *ConnectivityMatrix) Lookup(src, dst *Endpoint, pp ProtoPort) Connectivi
...
+			}
+			if pair.Destination != nil && pair.Destination.External != nil && extAllowBySource[pair.Source] {
+				continue
+			}
+			unknowns = append(unknowns, describeUnknownEntry(pair, pp, e))
+		}
+	}
+	if len(unknowns) > 0 {
+		slices.Sort(unknowns)
+		errs = append(errs, fmt.Errorf("%d matrix entr(ies) could not be evaluated and were not overlaid by the test: %s", //nolint:goerr113
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/matrix.go
```diff
@@ -332,9 +268,121 @@ func (m *ConnectivityMatrix) Lookup(src, dst *Endpoint, pp ProtoPort) Connectivi
...
+	}
+
+	extAllowBySource := map[*Endpoint]bool{}
+	for pair, byPP := range m.entries {
+		for _, e := range byPP {
+			if e.Verdict == VerdictAllow && pair.Destination != nil && pair.Destination.External != nil {
+				extAllowBySource[pair.Source] = true
+			}
+		}
+	}
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/endpoints.go
```diff
@@ -165,16 +156,26 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
...
+
+		// A multihomed hostBGP server has several VPCAttachments pointing at the
+		// same (vpc, subnet) — one per connection — but a single /32 VIP. Collapse
+		// them to one candidate attachment so the server yields exactly one
+		// endpoint instead of dropping the duplicates with misleading warnings.
+		if slices.ContainsFunc(serverAttachments[serverName], func(a serverAttachment) bool {
+			return a.vpcName == vpc.Name && a.subnetName == subnetName
+		}) {
+			continue
 		}
```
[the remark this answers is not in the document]
edipascale: addressed

pkg/hhfab/matrix.go
```diff
@@ -597,13 +608,146 @@ func runMatrixPortForwardPhase(ctx context.Context, opts TestConnectivityOpts, m
...
+		}
+		cmd := fmt.Sprintf("sudo docker exec -d iperf3 iperf3 -s -p %d", hp.port)
+		if _, stderr, err := retrySSHCmd(ctx, ssh, cmd, hp.host); err != nil {
+			teardown()
+
+			return nil, fmt.Errorf("starting proto-port iperf3 listener on %s:%d: %w: %s", hp.host, hp.port, err, stderr)
+		}
+		started = append(started, hp)
+		slog.Debug("Started proto-port iperf3 listener", "host", hp.host, "port", hp.port)
+	}
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/rt_acl_tests.go
```diff
@@ -0,0 +1,489 @@
...
+				Rules: []gwapi.PeeringACLRule{
+					// forward: select the subnets by VPC subnet name
+					{
+						Name: "allow-subnet-fwd", From: vpc1.Name, To: vpc2.Name,
+						Action: gwapi.ACLActionAllow, Scope: gwapi.ACLScopePacket,
+						Match: gwapi.PeeringACLMatch{
+							Source:      []gwapi.PeeringACLMatchEndpoint{{VPCSubnet: "subnet-01"}},
+							Destination: []gwapi.PeeringACLMatchEndpoint{{VPCSubnet: "subnet-01"}},
+						},
+					},
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/matrix.go
```diff
@@ -720,11 +917,21 @@ func runMatrixIperfPortForward(ctx context.Context, opts TestConnectivityOpts, i
...
+			return &IperfError{
+				Source:      from,
+				Destination: target,
+				Why:         why,
+				ClientMsg:   "port-forward target is reachable but the matrix expects it to be denied",
+			}
+		}
+
+		return nil
+	}
```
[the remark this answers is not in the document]
edipascale: fixed

pkg/hhfab/rt_acl_tests.md
```diff
@@ -0,0 +1,86 @@
...
+| Default action `deny` / `deny-unless-exposed` | Default Deny / Deny-Unless-Exposed |
+| Rule action `allow` / `deny` | all allow cases / carve-out + precedence |
+| Protocol `tcp` / `udp` / any | Protocol Scoping, Port Range / UDP Carve-Out / Explicit Allow, Subnet |
+| Selector `VPCSubnet` / `CIDR` | Subnet/CIDR Scoping (both, one per direction) |
+| Ports single / range, dst / src side | Precedence (single dst+src) / Port Range (range dst+src) |
+| Scope `packet` / `flow` | all packet cases / Flow Scope Masquerade |
+| Rule precedence (first-match) | Precedence Allow-Then-Deny + Deny-Then-Allow |
+| Stateless return-path requirement | Packet One-Way (negative) |
+
+### Known limitations / assumptions
```
pau-hedgehog, line 80: Checking githedgehog/docs#338, there are two possible gaps.

The docs say ACL rules are evaluated before any NAT, so match.dst is compared against the address the initiator dialed (the expose's "as"), and that's the first worked example. No case exercises it: nine configure no NAT, and the masquerade one uses a rule with no match block, so nothing ever matches a dst against an as-advertised address.

The docs also say proto takes tcp, udp, or a quoted protocol number, and that "1" is currently the only way to match ICMP (an icmp keyword is future work). So every ICMP expectation rests on ICMP falling through to the default action, and nothing asserts that a rule can match ICMP at all.

Is it worth adding a case for each?

pkg/hhfab/matrix.go
```diff
@@ -332,9 +275,162 @@ func (m *ConnectivityMatrix) Lookup(src, dst *Endpoint, pp ProtoPort) Connectivi
...
+
+	return errors.Join(errs...)
+}
+
+// defaultEntryShadowed reports whether a default-ProtoPort entry is read by no
+// phase at all: runMatrixServerServerPhase hands a pair carrying proto-port
+// entries to runMatrixProtoPortPhase, which reads only those. Port-forward
+// entries are exempt, runMatrixPortForwardPhase still reads them.
+func (m *ConnectivityMatrix) defaultEntryShadowed(pair EndpointPair, pp ProtoPort, e ConnectivityExpectation) bool {
```
pau-hedgehog, line 334: Three separate places now decide which phase owns an entry:

- runMatrixServerServerPhase: skips a pair if HasProtoPortEntries, or if NAT.DestinationPort is set
- runMatrixPortForwardPhase: claims an entry if NAT is set, the verdict is Allow or Deny, the IP is valid and the port is non-zero
- this function: decides when a default entry is read by nobody

They agree today, and treating any non-zero DestinationPort as claimed is the right call, since it errs toward flagging. The concern is drift. This function only exists because Validate ended up stricter than the phases, and the next edit to either gate can reintroduce exactly that.

Would one entryOwner(pair, pp, e) helper, consulted by both the phases and Validate, be worth it? A follow-up is fine.

pkg/hhfab/matrix.go
```diff
@@ -597,13 +644,149 @@ func runMatrixPortForwardPhase(ctx context.Context, opts TestConnectivityOpts, m
...
+				case "icmp":
+					deps.wg.Go(func() {
+						if pe := checkPing(ctx, opts.PingsCount, deps.pings, fromName, toName, fromSSH, toIP, nil, expected); pe != nil {
+							deps.errChan <- pe
+						}
+					})
+				case "tcp":
+					port := pp.Port
+					deps.wg.Go(func() {
+						if ie := checkTCPPort(ctx, deps.iperfs, fromName, fromSSH, toIP, port, expected.Reachable); ie != nil {
```
pau-hedgehog, line 761: checkTCPPort and checkUDPPort take a bool rather than a Reachability, so neither ever sets IperfError.Why. The effect is that the ten ACL cases are the only failures in the suite that won't carry the new "because: ..." suffix.

runMatrixIperfPortForward works around it by setting ie.Why after the call returns. That reads like the signature is the problem rather than the caller. Taking Reachability, as checkPing and checkCurl now do, would cover both.

pkg/hhfab/testing.go
```diff
@@ -3315,13 +3379,180 @@ func isReportableIperfError(ie *IperfError) bool {
...
+
+			return ie
+		}
+
+		return nil
+	}
+
+	// expected == deny.
+	if !blocked {
+		ie.ClientMsg = fmt.Sprintf("should not be reachable but UDP datagrams delivered (packets %d, loss %.1f%%)", packets, lostPercent)
```
pau-hedgehog, line 3540: Loss between the two thresholds, 90% and 99%, counts as neither delivered nor blocked, so an expected-allow and an expected-deny both fail there. That's the right behavior.

The message doesn't reflect it. For an expected deny it prints "should not be reachable but UDP datagrams delivered (packets N, loss 95.0%)", which reads as a clean observation of delivery and could send someone hunting a dataplane bug when the path is simply lossy.

pkg/hhfab/rt_acl_tests.md
pau-hedgehog, line 80: Two more things, for completeness:

ACLs on external peerings can't be expressed yet. docs#338 allows externals in a rule's from/to (prefixed ext.) and its second worked example is an egress allow-list towards one, but runMatrixProtoPortPhase skips any pair whose destination isn't a server, so a ProtoPort entry on a server-to-external pair gets probed by nothing. Validate doesn't catch it either, since it only inspects Unknown verdicts. An external ACL case would report green while asserting nothing, which is worth knowing before we try.

Second, the flow-scope note is accurate but reads as permanent, while docs#338 describes it as temporary. A link would tell the next reader it's a moving target, and it's the trigger for revisiting the packet-everywhere choice: once flow scope works without NAT it becomes the CRD default path with no coverage.

pkg/hhfab/rt_acl_tests.go
```diff
@@ -0,0 +1,431 @@
...
+	"fmt"
+
+	gwapi "go.githedgehog.com/fabric/api/gateway/v1alpha1"
+	vpcapi "go.githedgehog.com/fabric/api/vpc/v1beta1"
+)
+
+const (
+	// aclProbePort is served by the always-on iperf3 daemon (TCP+UDP), so
+	// tests probing it need no on-demand listener.
+	aclProbePort    uint16 = 5201
```
pau-hedgehog, line 17: nit: duplicates `persistentIperf3Port` from matrix.go

pkg/hhfab/vlabbuilder.go
```diff
@@ -376,6 +377,10 @@ func (b *VLABBuilderDefault) Build(ctx context.Context, l *apiutil.Loader, fabri
 		for eslagLeafID := uint8(0); eslagLeafID < leafs; eslagLeafID++ {
 			leafName := fmt.Sprintf("leaf-%02d", leafID+eslagLeafID)
 			leafNames = append(leafNames, leafName)
+			// add eslag leaves to candidates for multihomed servers if there are not enough orphan laves
+			if b.OrphanLeafsCount < 2 {
+				mhLeaves = append(mhLeaves, leafName)
```
pau-hedgehog, line 382: nit: ESLAG leaves are appended before the orphan loop, so with one orphan the first mh server takes two ESLAG leaves and the orphan is never used, which contradicts "preferably orphans"

pkg/hhfab/testing.go
```diff
@@ -425,38 +425,59 @@ func GetServerNetconfCmd(conn *wiringapi.Connection, opts ServerNetconfOpts) (st
...
-	}
 
-	cmd += strings.Join(interfaces, ":i=")
-	addr := subnet.Addr()
-	for range serversInSubnet {
-		addr = addr.Next()
-	}
-	if !addr.IsValid() {
-		return "", fmt.Errorf("failed to get IP address from subnet %s", subnet.String())
+		cmd += fmt.Sprintf("%s:v=%d:i=", param.VPCLabel, param.VLAN)
```
pau-hedgehog, line 470: 555d3abc changes the first argv token from the literal "vpc" to the VPC's name, and that string goes straight through as arguments to the host-bgp container. Does host-bgp:v0.4.0 treat the leading token as an opaque label? The v0.4.0 "vlan alias" note suggests yes. Not sure where the format is documented

Since getServerHostBGPCmd is a pure function with no test today, is it worht adding one? A single-param and a two-param assertion on the emitted string would pin the new format and cover the []HostBGPParams loop, which has no caller until #1747

pau-hedgehog (approved):
Nice work, the oracle part specially: Validate() plus VerdictUnknown closes the pass-by-testing-nothing failure mode, as we saw on CI last week.

I checked ACL enforcement in the dataplane repo, so the ten cases assert what they claim. The "expected to fail" in one of the commits is stale. The rest of my comments are small.

Next, multihomed/hostBGP on #1747 which Github is suggesting to stack. Do we want that?

pkg/hhfab/matrix.go
```diff
@@ -597,13 +644,149 @@ func runMatrixPortForwardPhase(ctx context.Context, opts TestConnectivityOpts, m
...
+		defer cancel()
+		for _, hp := range started {
+			ssh := deps.sshByServer[hp.host]
+			if ssh == nil {
+				continue
+			}
+			cmd := fmt.Sprintf("sudo docker exec iperf3 pkill -f 'iperf3 -s -p %d'", hp.port)
+			if _, stderr, err := retrySSHCmd(tctx, ssh, cmd, hp.host); err != nil {
+				slog.Warn("Failed to stop proto-port iperf3 listener", "host", hp.host, "port", hp.port, "err", err, "stderr", stderr)
+			}
```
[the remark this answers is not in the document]
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/matrix.go
```diff
@@ -597,13 +644,149 @@ func runMatrixPortForwardPhase(ctx context.Context, opts TestConnectivityOpts, m
...
+					continue
+				}
+
+				switch pp.Protocol {
+				case "icmp":
+					deps.wg.Go(func() {
+						if pe := checkPing(ctx, opts.PingsCount, deps.pings, fromName, toName, fromSSH, toIP, nil, expected); pe != nil {
+							deps.errChan <- pe
+						}
+					})
```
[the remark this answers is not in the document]
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/rt_acl_tests.md
edipascale: added those two cases in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/matrix.go
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933; claude simplified that to `entryOwner(e)` as the other two fields were always taken from the ewntry anyway

pkg/hhfab/matrix.go
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/testing.go
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/rt_acl_tests.md
edipascale: addressed both in https://github.com/githedgehog/fabricator/pull/1933; specifically re the first issue we are now erroring at the validate phase, we can remove this once we have a way to test this

pkg/hhfab/rt_acl_tests.go
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/vlabbuilder.go
edipascale: addressed in https://github.com/githedgehog/fabricator/pull/1933

pkg/hhfab/testing.go
edipascale: I can confirm the first token is just an arbitrary word that is used for route-maps associated with that VPC subnet. added a unit-test for the command helper in https://github.com/githedgehog/fabricator/pull/1933

Frostman (approved):

PR #1911, opened 2026-07-29
fix(hhfab): stop hardened sshd config from overriding VLAB SSH limits

### LINKED ISSUE #1912: VLAB show-tech intermittently fails with ssh handshake failed EOF
hhfab vlab show-tech fails to collect diagnostics from a subset of switches/servers on every run, with errors like:
```
collecting show-tech for s5248-02: uploading script to s5248-02: creating sftp: connecting: ssh: handshake failed: EOF
```

All VLAB switch and server SSH access is proxied through the control node's sshd (core@127.0.0.1 on the forwarded port). show-tech launches connections to every device concurrently, and this is the only source address that sshd on the control node sees for all of them.

#1528 raised that sshd's MaxStartups/MaxSessions via 99-vlab.conf specifically to allow this concurrency. #1843 later added a hardening baseline at 40-hedgehog.conf that sets its own MaxStartups and, new, PerSourceMaxStartups 4. Since sshd_config's Include "*.conf" expands alphabetically and sshd keeps the first value it sees per keyword, 40-hedgehog.conf (sorted before 99-) silently wins over 99-vlab.conf's override, reinstating the original connection-limit failure plus a stricter PerSourceMaxStartups cap neither PR accounted for

pau-hedgehog:
## Summary

VLAB switch/server SSH always proxies through the control node's sshd (`core@127.0.0.1` on the forwarded port). #1528 raised that sshd's `MaxStartups`/`MaxSessions` via `99-vlab.conf` to stop `show-tech` bursts from hitting connection limits and failing with `ssh: handshake failed: EOF`.

#1843 added a hardening baseline at `40-hedgehog.conf` that sets its own `MaxStartups` and, new, `PerSourceMaxStartups 4`. Since `sshd_config`'s `Include "*.conf"` expands alphabetically and sshd keeps the *first* value it sees per keyword, `40-hedgehog.conf` (before `99-` in sort order) silently wins over `99-vlab.conf`, reinstating the original bug plus a stricter failure mode neither PR was aware of.

Fixes #1912

### DIFF
--- a/pkg/hhfab/vlabrunner.go
+++ b/pkg/hhfab/vlabrunner.go
@@ -1124,10 +1124,17 @@ func (c *Config) vmPostProcess(ctx context.Context, vlab *VLAB, d *artificer.Dow
 		slog.Debug("Node install marker is complete", "vm", vm.Name, "type", vm.Type)
 
 		if vm.Type == VMTypeControl {
-			// Increase SSH connection limits on the control node to prevent errors during show-tech
-			// Only needed for VLAB testing, not in production where the default limits provide DoS protection
+			// Increase SSH connection limits on the control node to prevent errors during show-tech.
+			// Only needed for VLAB testing, not in production where the default limits provide DoS protection.
+			// Named 00- (not 99-) because sshd_config keeps the first value it sees for a repeated
+			// keyword: the hardening baseline in 40-hedgehog.conf sets its own MaxStartups and
+			// PerSourceMaxStartups, and being read first (alphabetically before 99-) it would otherwise
+			// silently win over this override. PerSourceMaxStartups also has to be relaxed explicitly
+			// since 40-hedgehog.conf is the only place that sets it; all VLAB switch/server access is
+			// proxied through this one control node connection, so a strict per-source cap throttles
+			// concurrent show-tech collection even with MaxStartups raised.
 			slog.Debug("Increasing SSH connection limits for VLAB", "vm", vm.Name)
-			sshConfigCmd := `bash -c 'echo -e "MaxStartups 100:30:200\nMaxSessions 50" | sudo tee /etc/ssh/sshd_config.d/99-vlab.conf > /dev/null && sudo systemctl restart sshd'`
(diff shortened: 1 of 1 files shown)


pkg/hhfab/vlabrunner.go
```diff
@@ -1124,10 +1124,17 @@ func (c *Config) vmPostProcess(ctx context.Context, vlab *VLAB, d *artificer.Dow
...
+			// Named 00- (not 99-) because sshd_config keeps the first value it sees for a repeated
+			// keyword: the hardening baseline in 40-hedgehog.conf sets its own MaxStartups and
+			// PerSourceMaxStartups, and being read first (alphabetically before 99-) it would otherwise
+			// silently win over this override. PerSourceMaxStartups also has to be relaxed explicitly
+			// since 40-hedgehog.conf is the only place that sets it; all VLAB switch/server access is
+			// proxied through this one control node connection, so a strict per-source cap throttles
+			// concurrent show-tech collection even with MaxStartups raised.
 			slog.Debug("Increasing SSH connection limits for VLAB", "vm", vm.Name)
-			sshConfigCmd := `bash -c 'echo -e "MaxStartups 100:30:200\nMaxSessions 50" | sudo tee /etc/ssh/sshd_config.d/99-vlab.conf > /dev/null && sudo systemctl restart sshd'`
+			sshConfigCmd := `bash -c 'echo -e "MaxStartups 100:30:200\nMaxSessions 50\nPerSourceMaxStartups none" | sudo tee /etc/ssh/sshd_config.d/00-vlab.conf > /dev/null && sudo rm -f /etc/ssh/sshd_config.d/99-vlab.conf && sudo systemctl restart sshd'`
```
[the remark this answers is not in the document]
pau-hedgehog: verified it does not apply to flatcar

Frostman, line 1137: There is no need to delete the `99-` one, it'll not be present on new vlabs.

mrbojangles3, line 1137: @pau-hedgehog Do we know how many sessions and maxstartups we use? or do we just know what settings work for us? 

I ask because @Frostman and I had a call and figured if we were hitting an issue in CI then we might hit the issue in production. So I am just wondering if we know our usage or just a high water mark. The end goal might be merge the configs.

We can get rid of PerSourceMaxStartups, since it can be covered by MaxStartups. I am preparing a PR to be less strict and will integrate your feedback

Frostman (changes requested):
We' discussed with @mrbojangles3 that we'll bump our hardened ssh config with values close to it/

PR #1913, opened 2026-07-29
fix(rt): honor gateway expose Not in the connectivity matrix

### LINKED ISSUE #1906: Release-test expectation model ignores gateway-expose Not exclusions (order-dependent)
[`isVPCSubnetPresentInPeering`](https://github.com/githedgehog/fabricator/blob/e0bd7a3f9/pkg/hhfab/testing.go#L2904) (the function that derives expected reachability across a gateway peering) handles expose `Not` exclusion entries inconsistently depending on their position in the `expose.IPs` list:

- include entry before the `Not` entry: returns `(true, nil)`, the exclusion is **silently ignored**, and every excluded address is expected reachable;
- `Not` entry before the include entry: returns a `reachCheckUnsupported` error.

Same semantic configuration, opposite verdicts, decided by list order. The gateway dataplane *does* honor `Not` (verified end-to-end, below), so whenever a test uses an exclusion, the expectation model over-approximates and the release test fails spuriously.

## Root cause

Per the gateway API, each `expose.IPs` entry carries exactly one of `cidr`/`not`/`vpcSubnet` ([`gatewaypeering_types.go:186-193`](https://github.com/githedgehog/fabric/blob/v0.126.0/api/gateway/v1alpha1/gatewaypeering_types.go#L186-L193), enforced by `Validate`): a `Not` exclusion is a standalone entry sitting alongside include entries. But `isVPCSubnetPresentInPeering` returns `true` on the first include entry matching the queried subnet ([`testing.go:2933-2935`](https://github.com/githedgehog/fabricator/blob/e0bd7a3f9/pkg/hhfab/testing.go#L2933-L2935)), never scanning the rest of the list, and routes `Not`-only entries to the error branch ([`testing.go:2929-2930`](https://github.com/githedgehog/fabricator/blob/e0bd7a3f9/pkg/hhfab/testing.go#L2929-L2930)) only if the loop reaches them first. Structurally, the function's per-subnet boolean return type cannot express "reachable except this range" at all.

## Reproduction (VLAB, mesh/l2vni)

Branch `pau/rt-gw-peering-expose-not` @ `e0bd7a3f9` (based on [fabricator#1761](https://github.com/githedgehog/fabricator/pull/1761), the connectivity-matrix refactor) adds `Gateway Peering Expose Not` to the Multi-VPC Single-Subnet suite: a gateway peering between vpc-03 (server-3) and vpc-04 (server-4), plus a `Not` entry that excludes vpc-04's own server address (`10.10.4.2/32`) from vpc-04's expose ([`gatewayPeeringExposeNotTest`](https://github.com/githedgehog/fabricator/blob/e0bd7a3f9/pkg/hhfab/rt_multi_vpc_single_subnet_suite.go#L357-L405)).

Run `20260729-003035-release-test-default-default-e0bd7a3f9-pau` (env-1, mesh, l2vni, wiring `f9dedea`). Setup, `release-test.log:2829-2830`:

```
01:51:22 INF Excluding server address from gateway peering expose server=server-4 vpc=vpc-04 not=10.10.4.2/32
01:51:22 INF Enforcing GatewayPeering name=vpc-03--vpc-04 vpc1=vpc-03 vpc2=vpc-04
```

Result, `release-test.log:2927` (one line; the `\n` inside the `error` field's quoted string is the literal two characters logged by slog, not a line break):

```
01:51:47 ERR FAIL test="Gateway Peering Expose Not" error="testing gateway peering connectivity with excluded 10.10.4.2/32: ping server-4 -> server-3: should be reachable but ping failed (sent 5, rcvd 0, expected true)\nping server-3 -> server-4: should be reachable but ping failed (sent 5, rcvd 0, expected true)"
```

Reading: for `server-3 -> server-4` (destination is the excluded `/32` itself), the expectation model ignored the exclusion (include entry listed first), expected Allow, and the dataplane correctly denied it: a genuine deny, not a probe failure. This confirms the dataplane implements `Not` and that half of the divergence is entirely in the hhfab expectation model.

`server-4 -> server-3` (destination *not* excluded) also failed 5/5. This direction is not explained by the expectation-model bug above, and its root cause is not established. The connectivity-matrix summary (`ping=2 iperf=0 curl=0`, `release-test.log:2926`) shows only these two pings failed in the whole matrix, nothing else was affected.

Next step to narrow this down: exclude an unused address in vpc-04 (e.g. `10.10.4.99/32`) instead of the server's own address, and rerun. If server-3/server-4 connectivity stays intact, the effect is scoped to traffic touching the excluded address; if it still fails, something broader is affected by the exclusion.

## Impact

- Any future suite (or user-driven test) using expose exclusions gets spurious release failures, or silently skipped pairs, depending on YAML entry order.
- The path is currently unexercised: no existing suite constructs a `Not` entry, which is why this has never fired in CI.
- The connectivity-matrix refactor ([fabricator#1761](https://github.com/githedgehog/fabricator/pull/1761)) inherits the bug unchanged: `populateConnectivityMatrix` calls the same `isVPCSubnetPresentInPeering` shown above.

## Suggested direction

The matrix introduced by [fabricator#1761](https://github.com/githedgehog/fabricator/pull/1761) is the natural place for a fix: matrix entries are per endpoint pair, so an exclusion maps to per-pair Deny entries for servers whose IP falls inside `Not`, the same shape the NAT overlays already use. Either an exclusion-aware overlay or a corrected shared helper (scan the full entry list; make the verdict order-independent) would do.

pau-hedgehog:
Fixes #1906.

`isVPCSubnetPresentInPeering` handled expose `Not` entries inconsistently depending on their position in `expose.IPs`: an include listed first silently ignored the exclusion, an exclusion listed first returned `reachCheckUnsupported`. Same configuration, opposite verdicts, decided by list order.

Per the gateway API a `Not` is a standalone entry alongside the includes, and it narrows an already-exposed range down to individual addresses, which a per-subnet boolean cannot express. The helper now scans every entry of an expose before answering, so its verdict no longer depends on order, and it returns that expose's `Not` ranges alongside the verdict in `Reachability.SourceExclusions` and `DestExclusions`. Callers that know the endpoint addresses apply them: `populateConnectivityMatrix` records Deny at the point it builds the expectation, and `TestConnectivity` narrows its own verdict with the server addresses it already discovered, so a `Not` is covered on the non-matrix path as well. Both directions are denied, matching the dataplane: traffic addressed to an excluded host is dropped, and the excluded host's own traffic has no return path across the peering.

An expose entry setting none of cidr, not or vpcSubnet is a malformed peering that admission rejects, so it now returns a plain error instead of `reachCheckUnsupported`.

Also adds two release-test cases. `Gateway Peering Expose Not` peers vpc-03/vpc-04 and excludes vpc-04's own server address from its expose. `Gateway Peering Expose Not Unused` is the same setup but excludes an address nothing uses, derived programmatically from the subnet and checked against every discovered endpoint IP. The second case is kept because on a topology with one server per VPC the first one has no other endpoint left in that VPC to probe, so it cannot distinguish the excluded host being correctly denied from the whole exposed subnet going dark.

Test plan (manual, env-1, mesh/l2vni, via env-ctl, run `20260729-224422-release-test-default-default-fffd13537-pau`):

- Release test PASS 41/47, 6 skipped, 0 failed.
- `Gateway Peering Expose Not` passes. It failed before this change, with the matrix expecting Allow and the dataplane denying.
- `Gateway Peering Expose Not Unused` passes, so the exclusion stays host-scoped.

That run predates both the move of the exclusion check into the reachability layer and the rebase onto current master, so it validates the expected verdicts but not the code that now produces them. A re-run is pending before merge.


pau-hedgehog:
@coderabbitai review

pau-hedgehog:
The earlier reviews ran while this branch still carried pre-merge copies of #1761's commits, so the diff showed 19 files and most findings landed on code that is now merged in master. Rebased onto master, the diff is 4 files and one finding was in scope.

Fixed: `pickUnusedHostAddress` computed its loop bound as `uint32(1) << (32 - bits)`, which wraps to 0 for a `/0` and underflows the bound to 4294967295. The bound is now computed in uint64 while the loop counter stays uint32, so gosec's G115 stays quiet. Added `TestPickUnusedHostAddress` covering `/0`, `/31`, IPv6, and the used-address skip.

Out of scope, all in code merged with #1761 rather than added here: the `noSetup` endpoint-less matrix in `rt_base.go`, `findAllServersInSubnet` native-VLAN handling in `rt_utils.go`, the unused `matrix` parameters in `rt_single_vpc_suite.go`, `toolboxMutexes` in `matrix.go`, and memoizing `IsServerReachable`. The `noSetup` and native-VLAN ones look worth their own issues.

One correction: the suggestion that `%s` on `expose.As` renders as `%!s(...)` is wrong. `PeeringEntryAs` has only string fields, so `%s` on the slice prints `[{10.0.0.0/24 } { 10.0.0.5/32}]`. That line is also from #1761, not this PR.

pau-hedgehog:
On the IPv4 subnet selection: `pickUnusedHostAddress` returns an error for any prefix where `prefix.Addr().Is4()` is false (`rt_utils.go:828`), so it cannot return an IPv6 address and no IPv6 `/32` exclusion can be constructed. An IPv6 first-sorted subnet would surface as a clear error from that call, not as a broad exclusion. Leaving the selection as is.

On normalizing `used`: taking this one, though the mismatch was upstream of the lookup. The `!ep.Server.IP.Is4()` filter returns false for 4-in-6 mapped addresses, so such an endpoint was dropped before it ever entered the set, and probing both forms inside the loop would not have helped. Now unmapping when building the set, which also matches `isEndpointExcluded`, which already unmaps before its prefix check.

pkg/hhfab/testing.go
```diff
@@ -2906,30 +2921,118 @@ func isVPCSubnetPresentInPeering(peering *gwapi.PeeringEntry, vpc gwapi.VPCInfo,
...
 					if subnet.CIDR == exposeEntry.CIDR {
 						exposeSubnetName = subnetName
 					}
 				}
-			} else {
-				return false, fmt.Errorf("%w: gw peering with non-empty expose 'not' %s in IPs", reachCheckUnsupported, exposeEntry.Not)
+			case exposeEntry.Not != "":
+				continue
+			default:
+				return false, fmt.Errorf("%w: gw peering expose entry with none of cidr, not, vpcSubnet set", reachCheckUnsupported)
```
edipascale: `reachCheckUnsupported` was meant to mean "this is a valid question but the reachability check does not support it yet"; what we have here however is a malformed peering, which with our current validations should be unreachable. I'd just return a normal error

pkg/hhfab/testing.go
```diff
@@ -2906,30 +2921,118 @@ func isVPCSubnetPresentInPeering(peering *gwapi.PeeringEntry, vpc gwapi.VPCInfo,
...
+			case exposeEntry.CIDR != "":
 				for subnetName, subnet := range vpc.Spec.Subnets {
 					if subnet.CIDR == exposeEntry.CIDR {
 						exposeSubnetName = subnetName
 					}
 				}
-			} else {
-				return false, fmt.Errorf("%w: gw peering with non-empty expose 'not' %s in IPs", reachCheckUnsupported, exposeEntry.Not)
+			case exposeEntry.Not != "":
+				continue
```
edipascale, line 3081: this is not the correct thing to do; there's a `Not` statement and we are ignoring it, but still continuing and potentially returning a verdict, so we are pretending that we can correctly check whether the destination is reachable. We should either fix the check and evaluate the exclusion here (which would allow us to test `Not` statements without the extra machinery of the matrix connectivity checks) or at least return a `reachCheckUnsupported` error here so that, if you are not using the matrix, you'll correctly get an error

edipascale (changes requested):
I think it's great to add support for `Not` statements, but we need to make sure that the reach check stays honest, e.g. in case it is used from the kubectl plugin or with legacy test connectivity checks.

As a side note, it's easier for me to look at the PR if you answer the agent's comments directly (and ideally resolve them if it's the case) rather than as separate comments. I also think we should start doing a pass after an agent-aided PR and trim all unnecessary comments - I know I have been guilty of not doing this. Maybe there's a way to tune claude's verbosity with instructions in the CLAUDE.md and we should consider adding one, but this is a separate topic of course.

pkg/hhfab/testing.go
pau-hedgehog: Fixed the check instead: isVPCSubnetPresentInPeering now returns the matching expose's Not ranges alongside the verdict, and TestConnectivity applies them to the addresses it already discovered, so Not is testable without the matrix. That also let me delete the separate exclusion pass in populateConnectivityMatrix.

pkg/hhfab/testing.go
pau-hedgehog: Agreed, plain error now.

pkg/hhfab/testing.go
```diff
@@ -2741,43 +2901,138 @@ func IsSubnetReachableWithGatewayPeering(ctx context.Context, kube kclient.Reade
...
+// here is what makes the verdict independent of entry order.
 func isVPCSubnetPresentInPeering(peering *gwapi.PeeringEntry, vpc gwapi.VPCInfo, vpcName string, vpcSubnet string) (bool, error) {
 	for _, expose := range peering.Expose {
 		if expose.DefaultDestination {
 			return true, nil
 		}
 
 		if len(expose.As) > 0 {
-			return false, fmt.Errorf("expose as %s is not supported yet", expose.As)
+			return false, fmt.Errorf("%w: gw peering with non-empty expose 'As' %s", reachCheckUnsupported, expose.As)
```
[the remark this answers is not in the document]
pau-hedgehog: Fixed, %v.

pkg/hhfab/matrix.go
```diff
@@ -0,0 +1,773 @@
...
+		if _, ok := toolboxMutexes[name]; ok {
+			continue
+		}
+		ssh, ok := sshConfigs[name]
+		if !ok {
+			return fmt.Errorf("no ssh config for server %q referenced by matrix", name) //nolint:goerr113
+		}
+		sshByServer[name] = ssh
+		toolboxMutexes[name] = &sync.Mutex{}
+	}
```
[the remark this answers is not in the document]
pau-hedgehog: It compiles (the map is read as a dedup set), but the mutexes are never locked, so it is dead weight. That code is in master now, not in this diff; separate cleanup.

pkg/hhfab/rt_utils.go
```diff
@@ -1226,3 +1274,52 @@ func findAnyAttachedServer(ctx context.Context, kube kclient.Client) (*AttachedS
...
+		}
+
+		servers = append(servers, ServerWithInterface{
+			Name:      serverNames[0],
+			Interface: ifName,
+		})
+	}
+
+	return servers, nil
+}
```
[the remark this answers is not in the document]
pau-hedgehog: Real, but findAllServersInSubnet is in master now, not in this diff, and nothing in the release tests creates a NativeVLAN attachment. Separate fix.

pkg/hhfab/rt_multi_vpc_single_subnet_suite.go
```diff
@@ -337,6 +354,146 @@ func gatewayPeeringTest(ctx context.Context, testCtx *VPCPeeringTestCtx, matrix
...
+			continue
+		}
+		usedIPs[ep.Server.IP] = true
+	}
+
+	unused, err := pickUnusedHostAddress(prefix, usedIPs)
+	if err != nil {
+		return false, nil, fmt.Errorf("picking an unused address in subnet %s of VPC %s: %w", prefix, vpc2.Name, err)
+	}
+	excluded := unused.String() + "/32"
```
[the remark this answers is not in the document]
pau-hedgehog: VPC subnets belong to an IPv4Namespace, so there is no IPv6 subnet to skip, and pickUnusedHostAddress already rejects a non-IPv4 prefix with a clear error. Leaving as is.

pkg/hhfab/rt_multi_vpc_single_subnet_suite.go
```diff
@@ -333,6 +350,152 @@ func gatewayPeeringTest(ctx context.Context, testCtx *VPCPeeringTestCtx, matrix
...
+
+	// Pick the first discovered server attachment in vpc2 and exclude its address as a /32.
+	excluded := ""
+	var excludedServer string
+	for _, ep := range matrix.AllEndpoints {
+		if ep.Server == nil || ep.Server.VPC != vpc2.Name || !ep.Server.IP.Is4() {
+			continue
+		}
+		excluded = ep.Server.IP.String() + "/32"
+		excludedServer = ep.Server.Name
```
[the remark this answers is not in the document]
pau-hedgehog: Applied in 0281c3c9. `Endpoint.Server.IP` comes from `discoverServerIPs`, which parses the dotted-quad output of `ip -o -4 addr show` with `netip.ParsePrefix`, so the address is always 4-byte today and `Unmap` is a no-op. Taking it for consistency with `gatewayPeeringExposeNotUnusedTest`, which already unmaps before the same check.

pau-hedgehog:
> I also think we should start doing a pass after an agent-aided PR and trim all unnecessary comments - I know I have been guilty of not doing this. Maybe there's a way to tune claude's verbosity with instructions in the CLAUDE.md and we should consider adding one, but this is a separate topic of course.

I have addressed this. I will take into account all your (as usual) good points from now on

pkg/hhfab/rt_utils.go
```diff
@@ -818,18 +823,51 @@ func buildNATConfig(mode NATMode, portForwardRules []gwapi.PeeringNATPortForward
 	return nat, nil
 }
 
+// pickUnusedHostAddress returns the first host address inside prefix that is not present
+// in used, skipping the network and broadcast addresses. IPv4 only.
+func pickUnusedHostAddress(prefix netip.Prefix, used map[netip.Addr]bool) (netip.Addr, error) {
```
edipascale, line 828: pickUnusedHostAddress returns the subnet gateway. It returns the first host address not claimed by a discovered endpoint, and used is seeded only from matrix.AllEndpoints. VPC subnet gateways default to the first IP in the subnet (vpc_types.go:93), and servers get .2, so it always returns .1. From the CI log:

```
23:19:11 INF Excluding an unused address from gateway peering expose vpc=vpc-08 subnet=10.0.8.0/24 not=10.0.8.1/32
```

10.0.8.1 is vpc-08's anycast gateway on the leaf, not an unused address. The test still passes and still exercises the prefix-fragmentation question it was written for, so this isn't a correctness bug — but the test name, the log line and the doc comment all claim "an address nothing uses", and that's not what it picks. It also means the test's signal is entangled with however the dataplane treats the SVI address, which is a bad thing to have quietly baked into a regression test.

Fix is one line in gatewayPeeringExposeNotUnusedTest: seed usedIPs with vpc2.Spec.Subnets[subnetNames[0]].Gateway (falling back to the subnet's first address when the field is empty) before calling pickUnusedHostAddress.

edipascale (approved):
Thanks @pau-hedgehog, LGTM. I asked Claude to have a look too and it found something, but it's a nit rather than a blocker

Frostman (approved):

pkg/hhfab/rt_utils.go
Frostman: @pau-hedgehog plz take care of it in a separate PR

Issue #1929, opened 2026-08-01
Full Mesh All Externals: single ICMP packet lost from route churn on a leaf outside the ping path


pau-hedgehog:
This looks similar in symptom shape to #1807 (single ICMP packet lost, route churn tied to a VPCPeering rollout) but the topology and the evidence trail differ enough that I'm filing it separately rather than assuming the same root cause. #1807 pins the cause to the endpoint leaf itself still applying VRF config at probe time. Here, both leaves that actually carry this probe's traffic had already finished applying, and the churn traces (not fully proven, see below) to a third leaf outside the ping path.

Run 30667616659, job 91279095067, variant v-up26.01-mesh-gw-iso-l2vni-rt (mesh topology, upgrade), build v0.48.0-11-g8eb81e9b, PR #1928 merge commit 8eb81e9b48dfe167539d880d07c74cdeb75ae224.

## Symptom

job.log:

```
2026-07-31T22:43:27.0927240Z 22:43:27 INF * Running test test="Full Mesh All Externals"
2026-07-31T22:48:37.2711643Z 22:48:37 ERR Test connectivity (matrix) failed ping=1 iperf=0 curl=0 took=2m47.71974487s errors="ping server-05 -> server-04: unexpected ping result (sent 5, rcvd 4, lost=[5], expected true)"
2026-07-31T22:48:37.2712512Z 22:48:37 ERR FAIL test="Full Mesh All Externals" error="ping server-05 -> server-04: unexpected ping result (sent 5, rcvd 4, lost=[5], expected true)"
```

Only the 5th (last) of 5 ICMP packets was lost. This harness does not log a per-packet DBG line with individual RTTs (unlike #1807's job.log), so there is no RTT-escalation curve for this occurrence, only the aggregate result.

## Path

job.log:

```
2026-07-31T22:09:57.7062842Z 22:09:57 INF Created vpcattachment=server-05--eslag--leaf-03--leaf-04--vpc-03--subnet-01
2026-07-31T22:42:05.3969663Z 22:42:05 INF Created vpcattachment=server-05--eslag--leaf-03--leaf-04--vpc-05--subnet-01
2026-07-31T22:43:17.5667875Z 22:43:17 INF Configured server=server-05 addr=10.0.5.2/24 conf="bond 1005 layer2+3 enp2s1 enp2s2"
[...]

PR #1933, opened 2026-08-03
More connectivity matrix fixes

edipascale:


pkg/hhfab/matrix.go
```diff
@@ -658,10 +694,10 @@ func startMatrixProtoPortListeners(ctx context.Context, matrix *ConnectivityMatr
 			continue
 		}
 		for _, dst := range matrix.AllEndpoints {
```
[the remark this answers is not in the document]
edipascale: addressed

pkg/hhfab/rt_eslag_fallback.go
```diff
@@ -87,7 +87,14 @@ func eslagFallbackTest(ctx context.Context, testCtx *VPCPeeringTestCtx, _ *Conne
 			return false, reverts, fmt.Errorf("building netconf command for %s: %w", server, err)
 		}
 		reverts = append(reverts, func(ctx context.Context) error {
-			return restoreServerBond(ctx, ssh, netconfCmd)
+			if err := restoreServerBond(ctx, ssh, netconfCmd); err != nil {
+				return fmt.Errorf("restoring server bond for %s: %w", server, err)
+			}
+			if err := testCtx.rebindMatrixServerEndpoint(ctx, matrix, server); err != nil {
```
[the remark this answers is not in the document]
edipascale: this is false

pkg/hhfab/matrix.go
```diff
@@ -304,45 +304,95 @@ func (m *ConnectivityMatrix) Validate() error {
...
 				continue
 			}
-			if pair.Destination != nil && pair.Destination.External != nil && extAllowBySource[pair.Source] {
+			if owner == matrixPhaseSkipped || e.Verdict != VerdictUnknown {
 				continue
 			}
-			if m.defaultEntryShadowed(pair, pp, e) {
+			if e.Pair.Destination != nil && e.Pair.Destination.External != nil && extAllowBySource[e.Pair.Source] {
 				continue
 			}
```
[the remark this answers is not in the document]
edipascale: addressed

pkg/hhfab/matrix.go
```diff
@@ -304,45 +304,95 @@ func (m *ConnectivityMatrix) Validate() error {
...
+	scoped := e.ProtoPort != (ProtoPort{})
 
-	return m.HasProtoPortEntries(pair.Source, pair.Destination)
+	switch {
+	case dst.Server != nil:
+		switch {
+		case scoped:
+			return matrixPhaseProtoPort
+		case portForward:
+			return matrixPhasePortForward
```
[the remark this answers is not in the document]
edipascale: addressed

pkg/hhfab/endpoints.go
```diff
@@ -183,6 +198,8 @@ func CollectServerEndpoints(ctx context.Context, kube kclient.Client, ssh SSHRes
 			subnetName: subnetName,
 			subnetCIDR: cidr,
 			hostBGP:    subnet.HostBGP,
+			eslag:      conn.Spec.ESLAG != nil,
+			l3vni:      vpc.Spec.Mode != vpcapi.VPCModeL2VNI,
```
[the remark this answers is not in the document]
edipascale: hostBGP subnets can only use unbundled connections, not ESLAG ones

pkg/hhfab/matrix.go
```diff
@@ -787,6 +845,14 @@ func (c *Config) TestConnectivityWithMatrix(ctx context.Context, vlab *VLAB, opt
 	if err := matrix.Validate(); err != nil {
 		return fmt.Errorf("connectivity matrix is not a sound oracle: %w", err)
 	}
+	// checkPing/checkIPerf/checkCurl no-op when their count is off, so a phase
+	// left with no enabled probe would pass its entries unasserted
+	if opts.PingsCount <= 0 && opts.IPerfsSeconds <= 0 && matrix.hasEntriesOwnedBy(matrixPhaseServerServer) {
```
pau-hedgehog: This guard asks hasEntriesOwnedBy, which iterates m.entries, so it only sees entries someone stored. runMatrixServerServerPhase doesn't work from stored entries though: it walks every endpoint pair and calls Lookup, which synthesizes a Deny for a pair with no stored entry. populate only stores reachable or unevaluable pairs, so a topology where nothing is reachable has zero stored server-to-server entries while the phase still probes all of them.

With --pings 0 --iperfs 0 that means the guard passes, the phase runs, checkPing returns nil immediately, and every deny expectation passes unasserted, which is the exact case the guard was added for. Counting over endpoint pairs (the same iteration the phase uses) rather than over stored entries would close it.

Not reachable from a release test or from the CLI today, so this is a hole in a safety net rather than a live bug.

pkg/hhfab/testing.go
```diff
@@ -3477,12 +3477,52 @@ const (
 	udpAllowLossThreshold = 90.0
 )
 
+const (
+	// A denied path neither completes nor refuses the iperf3 -u control connect,
+	// so iperf3 blocks in it until something bounds it. Deny probes get the short
+	// budget: on a path that is in fact open the handshake completes in
+	// milliseconds. Allow probes get a longer one so a momentarily slow but
+	// working path is not called down.
+	udpProbeDenyConnect  = 5 * time.Second
```
pau-hedgehog, line 3506: Sizing the deadlines off the connect budget is a nice shape, and the dependency holds: the toolbox image ships iperf 3.20+, and the command returns a parseable error report in 5.25s against a black-holed address.

The bit I'd push on is the asymmetry rather than the numbers. The probe is the oracle's only source of truth about the path, and its budget is now a function of what the oracle expects to find: a path that's genuinely open but slow to handshake gets less time precisely because the test expects it blocked, and a connect that would have completed at 6s reads as blocked, which satisfies the deny expectation. The failure is silent and biased towards agreeing with the test.

Whether 5s is enough is empirical and probably yes: I ran the ACL suite with a deliberately broken ACL on env-1 mesh and the deny probes behaved correctly through the 5s path. But RTTs of 46-308ms are on record in #1937's triage on these gateway paths, and that issue exists because packets there get lost in bursts. A lost SYN retransmits at ~1s, then 3s, then 7s cumulative, so two lost SYNs exceed a 5s budget.

A single budget for both polarities, 10 or 15s, costs about 10s per denied pair against the ~23s this commit already saves, and removes the coupling. Do you have measurements that make 5s the deliberate choice? Happy to leave it if so.

pkg/hhfab/matrix.go
```diff
@@ -750,6 +803,11 @@ func runMatrixProtoPortPhase(ctx context.Context, opts TestConnectivityOpts, mat
 
 				switch pp.Protocol {
 				case "icmp":
+					if opts.PingsCount <= 0 {
+						deps.errChan <- fmt.Errorf("matrix proto entry %s→%s expects icmp but pings are disabled", fromName, toName) //nolint:goerr113
```
pau-hedgehog: Two of the three disabled-probe checks are pre-flight in TestConnectivityWithMatrix and abort before anything runs. This one is inside runMatrixProtoPortPhase, so it fires after setup, listener start and every other probe in the phase, and it emits one error per entry rather than one per run. It's answerable from the matrix alone, so it could sit next to the other two.

pkg/hhfab/matrix.go
```diff
@@ -296,53 +296,123 @@ func (m *ConnectivityMatrix) Validate() error {
...
+	// nothing reads the entry by design: self-pair, or default entry superseded
+	// by the pair's proto-port entries
+	matrixPhaseSkipped
+	matrixPhaseServerServer
+	matrixPhasePortForward
+	matrixPhaseProtoPort
+	matrixPhaseCurl
+)
+
+func (m *ConnectivityMatrix) entryOwner(e ConnectivityExpectation) matrixPhase {
```
pau-hedgehog, line 352: Two things worth stating here, since this function is now the thing everyone will trust.

"Single source of truth for the phases' skip gates and for Validate" is very nearly right, but deps.inSources/inDestinations are a second gate Validate doesn't model, so an entry can be classified matrixPhaseProtoPort, satisfy Validate, and still never be probed because the --source/--destination filters excluded its pair. Narrowing the claim to ownership (rather than to whether the entry is probed) would keep it exact.

Second: dropping the port-forward phase's verdict guard is right, but it makes the phases depend on Validate having rejected Unknown first. reachabilityFromExpectation turns Unknown into Reachable:false, so a caller that reached a phase without Validate would assert deny on an entry explicitly marked unevaluable. Your commit message says this; the code doesn't.

pau-hedgehog (dismissed):
Nice work, this closes almost everything from the last round plus the two issues that came out of reviewing it.

One defect worth fixing: the new disabled-probe guard counts stored entries while the server-to-server phase probes synthesized ones, so it can pass while asserting nothing. One question on the UDP connect budget depending on the expectation. The rest are comments and one-liners.

I ran two mutations on env-1 to check the ACL suite fails when it should: breaking the ACL config fails Default Deny through all three probe types both ways, and flipping one tcp verdict fails exactly one probe in one direction. Worth knowing it has teeth, and maybe worth making permanent.

The red v-mesh-gw job looks like the #1929/#1937/#1924 single-packet-loss family: one lost ping on an expected-allow switch-peering pair, nothing to do with the ACL or matrix paths. Worth rebasing onto master before merge so CI re-runs and we get a clean board rather than merging with a red check.

Nits, batched here since several sit on lines this PR doesn't touch:

- overlayExternalNoEgress still writes a bare {Pair, Verdict: VerdictDeny} with no Reason, Peering or Detail, so a curl failure on one of those pairs prints "because: no reason recorded" in the branch whose parent added the because clause. One Detail line.
- The --multihomed-servers usage string still says "2 connections to 2 different orphan leaves" while 94fe6f2c makes ESLAG leaves the documented fallback. The struct comment was fixed in #1910, this one was missed.
- TestValidate is still in endpoints_test.go and this PR adds 40 more lines to it. Cheaper to move to matrix_test.go now than later.
- checkUDPPort's dead band: "UDP traffic was not blocked" is a real improvement over claiming delivery, but it still doesn't say the result was inconclusive or print the two thresholds.
- The UDP retry doubles the pathological worst case to 2 x outer, ~96s for a deny probe at release defaults, serialized behind the weight-1 semaphore. The expected case is far better and the retry is the right call; just noting the tail.

pkg/hhfab/testing.go
edipascale: I don't have measurements, I just wanted to make the test shorter when we could get away with it and it seemed like a solid round number. I'm happy to change it to something else if we think it's going to make the tests more robust, but not having observed any flake hitting this maybe we can wait?

pkg/hhfab/matrix.go
edipascale: addressed

pkg/hhfab/matrix.go
edipascale: addressed

pkg/hhfab/matrix.go
edipascale: addressed

edipascale:
@pau-hedgehog I rebased on master and pushed two more commits on top of the old ones; the first addresses your comments (except for the UDP connection timeout, until we decide what to do), the second improves the because clause for ACLs. I want to avoid a fourth round of matrix fixes if we can, so let's try and fix any remaining issue here

pkg/hhfab/testing.go
pau-hedgehog: Let's wait, agreed.

The failure needs two things at once: a deny expectation on a path that is genuinely open, which is a dataplane bug we would be masking rather than just a slow handshake, and that handshake then taking over 5s. And if it ever does bite it is diagnosable rather than silent, since reportErr already distinguishes "Connection timed out" from "Connection refused" and both land in the debug line.

So keep 5s. If we ever see a deny probe time out on a path we think should be open, that is the signal to revisit, and it is a one-constant change.

pau-hedgehog (approved):
Rebase looks good, and thanks for turning these around so fast.

checkProbesEnabled closes the guard properly: it walks the same pairs the phases do, including the Deny that Lookup synthesizes where populate stored nothing, and honouring the source/destination filters covers the ownership-is-not-probed gap as well. TestCheckProbesEnabled asserting the guard rather than the walk is the right shape.

Also reviewed 78e59df70 on top. expectationWhy keeps the allow path byte-identical and only changes the deny phrasing, TestExpectationWhy covers every branch of it, and each current caller of setVPCToVPCProtoVerdict is an ACL context so labelling a proto-scoped deny "ACL" holds for all of them. "ACL on gateway-peering ..." is a real improvement on a failure line, that phrasing bothered me in the mutation runs too.

UDP connect budget settled in-thread: keeping 5s, revisit if we ever see a deny probe time out on a path we expect to be open.

Nothing else from me.

Frostman (approved):

Issue #1834, opened 2026-06-19
ci: v-gw-onie-iso-* FAIL: control node ISO install stops after root partition resize, never reboots


pau-hedgehog:
https://github.com/githedgehog/fabricator/actions/runs/27685982028/job/81885715026

v-gw-onie-iso-l3vni, master aa21483a (v0.47.1-88-gaa21483a), 2026-06-17.

In --build-mode=iso mode, hhfab's dynamic SSH key is not present in authorized_keys on control-1 and gateway VMs after first-boot Ignition. Port 22 is listening (sshd socket-activated) but all auth attempts fail. hhfab waits the full 38-minute post-processing window, hits the context deadline, exits code 2. No tests ran.

Observed from serial logs:

- control-1: cat ~/.ssh/authorized_keys -> No such file or directory
- gateway-1: authorized_keys contains 2 unrelated static keys, not hhfab's key
- gateway-2: authorized_keys exists but is empty (header only)

Downstream: kubeconfig never downloaded; leaf switches stuck in ONIE with 169.254.x.x (link-local) only, fabric-dhcpd never served them; code 2 exit.

Not seen in the preceding successful run on the same runner (0-before artifact). Appears to be a timing or initialization race in key embedding during ISO build.

Next step: trace where the SSH pubkey is embedded into the ISO/Ignition config (pkg/fab/recipe/control_install.go, gateway ignition template) and verify it is finalized before the ISO is written/served.

pau-hedgehog:
Root cause is on control-1 alone: its ISO install stops after the root partition resize step and the node never reboots into the installed system, so it sits in the Flatcar live installer environment for the whole 40 minute post-processing window. The absent SSH key on control-1 is that live installer environment, which never holds hhfab's key. There is no first-boot Ignition failure here, and the gateways installed correctly with the expected keys. Every other symptom in the report is downstream of control-1.

control-1 state at collection time, 39 minutes after boot, from artifact fab-27685982028-v-gw-onie-iso-l3vni:

```
core@control-1 ~ $ uptime
 12:28:02 up 39 min,  2 users,  load average: 0.23, 0.10, 0.03
core@control-1 ~ $ df -h
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           2.9G   81M  2.9G   3% /
sysext          2.9G  8.0K  2.9G   1% /usr
overlay         2.9G   81M  2.9G   3% /etc
devtmpfs        4.0M     0  4.0M   0% /dev
[...]

PR #1960, opened 2026-08-09
chore(recipe): close the ignition file after writing it

pau-hedgehog:
The ignition file written into the install image was opened and never closed, leaking a file descriptor per installer built in ISO mode, where diskfs hands back a real os.File in the build workspace.

No behavior change. Close is a no-op in both backends of go-diskfs v1.4.2-hh3 (iso9660 sets a closed flag, fat32 drops its filesystem pointer), and durability already comes from O_SYNC in ISO mode and from the explicit fat32 Commit in USB mode. This just makes the ignition write match diskFSCopyFile a few lines below, which already defers Close on the same kind of handle.

Found while triaging #1834, unrelated to that failure.

### DIFF
--- a/pkg/fab/recipe/build.go
+++ b/pkg/fab/recipe/build.go
@@ -391,6 +391,8 @@ func buildUSBImage(ctx context.Context, opts buildInstallOpts) error {
 	if err != nil {
 		return fmt.Errorf("creating ignition file: %w", err)
 	}
+	defer ignFile.Close()
+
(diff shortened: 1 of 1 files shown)


mrbojangles3:
We are close to a RC / cutting a release. After we cut the release lets merge this.

mrbojangles3 (approved):

Frostman (approved):

PR #1971, opened 2026-08-11
fix(rt): attach multihomed servers in on-ready test

### LINKED ISSUE #1970: New VLAB OnReady Test fails RequireAllServers check when wiring includes a multihomed server
Job https://github.com/githedgehog/fabricator/actions/runs/31536723242/job/93931258412 failed the on-ready connectivity test, at commit 77c22a1f7dd659a505083a14d6736f8265d5e7e5 (tag v0.49.0).

The failure is deterministic and traces to a gap in the on-ready suite's handling of multihomed (unbundled, non-LACP) servers, not to a fabric or NOS problem.

job.log shows:

```
21:32:03 DBG Skipping server with unexpected connection count server=server-09 conns=2
21:33:14 ERR FAIL test="New VLAB OnReady Test" error="connectivity test failed: server \"server-09\" is not attached, but RequireAllServers is set"
```

server-09 is wired with two separate Unbundled connections (server-09--unbundled--leaf-01 and server-09--unbundled--leaf-03), the multihomed hostBGP-style pattern generated by vlab gen's --multihomed-servers flag.

pkg/hhfab/rt_on_ready_suite.go:120-124 skips any server whose connection count is not exactly 1 when deciding which servers to set up a VPCAttachment for:

```go
if len(conns.Items) != 1 {
    slog.Debug("Skipping server with unexpected connection count", "server", server.Name, "conns", len(conns.Items))

    continue
}
```

Because of this, server-09 never gets a VPCAttachment from the on-ready suite.

Separately, pkg/hhfab/testing.go:2437-2441 (the general TestConnectivity check, called with RequireAllServers=true whenever VPCMode is L2VNI, see pkg/hhfab/rt_base.go:98) walks every wiring server and requires each one to be attached:

```go
if sa, err := getServerAttachState(ctx, kube, &server, true); err != nil {
    return fmt.Errorf("checking server %q attachment state: %w", server.Name, err)
} else if !sa.Attached {
    if opts.RequireAllServers {
        return fmt.Errorf("server %q is not attached, but RequireAllServers is set", server.Name)
    }
```

server-09 has connections but no VPCAttachment, so sa.Attached is false, and the RequireAllServers check fails the whole test.

This server only shows up in the wiring because commit b76e09e54 ("ci: add one multihomed server for hostBGP") added --multihomed-servers=1 to vlab gen in .github/workflows/run-vlab.yaml for all run-vlab jobs, not only the hostBGP-specific suite. PR #1910 added multihomed-server support to SetupVPCs and TestConnectivity's endpoint discovery, but the on-ready suite's own server-selection filter above was never updated to match, so this now fails on every run-vlab job that includes a multihomed server in its wiring.

Fix options: either give multihomed servers a VPCAttachment in the on-ready suite's setup, matching what SetupVPCs already does per PR #1910, or exclude multihomed servers from the RequireAllServers check in TestConnectivity the same way ESLAG servers are excluded in non-L2VNI mode.

pau-hedgehog:
The on-ready suite skipped every server whose connection count was not exactly 1, so the multihomed server that `vlab gen --multihomed-servers=1` now adds to all run-vlab wirings never got a VPCAttachment, and `TestConnectivity` failed it under `RequireAllServers`.

Multihomed servers, meaning several connections all of which are unbundled, are now part of the discovered set, and the hostBGP VPC prefers one: it gets one VPCAttachment per link on the shared hostBGP subnet and all of its interfaces in the hostBGP container command, the same way `SetupVPCs` handles them. A multihomed server not picked for that role is attached over its first link.

## Validation

- [x] run-vlab l2vni job passes the on-ready suite
  - https://github.com/githedgehog/fabricator/actions/runs/31542840009
- [x] run-vlab l2vni job passes the release tests
  - https://github.com/githedgehog/fabricator/actions/runs/31548827308
 
Fixes #1970

### DIFF
--- a/pkg/hhfab/rt_on_ready_suite.go
+++ b/pkg/hhfab/rt_on_ready_suite.go
@@ -40,13 +40,17 @@ func makeOnReadyTestSuite() *JUnitTestSuite {
 	return suite
 }
 
-// ortServerInfo holds information about a server and its connection for the on-ready test.
+// ortServerInfo holds information about a server and its connections for the on-ready test.
 type ortServerInfo struct {
-	name        string
-	conn        wiringapi.Connection
-	switches    []string
-	isUnbundled bool
-	isMclag     bool
+	name string
+	// conn is the connection used when the server is attached over a single link
+	conn wiringapi.Connection
+	// conns holds every connection of the server, more than one only for multihomed ones
+	conns        []wiringapi.Connection
+	switches     []string
+	isUnbundled  bool
+	isMclag      bool
+	isMultihomed bool
 }
 
 // Test to get as much coverage as possible for Fabric while being short and compact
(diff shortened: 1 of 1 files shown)


pau-hedgehog:
@coderabbitai full review

Frostman (approved):

